PR #205, opened 2026-07-20
feat: cross-agent main-tree git guardrails

s205109:
Prevents local Claude Code, Codex, and GitHub Copilot sessions from mutating Git state in the human-owned main checkout, while leaving read-only Git and ordinary edit/build/test/format workflows there untouched.

Implements `docs/superpowers/plans/2026-07-20-cross-agent-worktree-guardrails-plan.md`.

## What lands

- **`scripts/agents/agent-worktree-guard.common.ps1`** — the single policy core: payload normalization, destructive-command rules, Git mutation detection, managed-worktree classification, bypass precedence, and messages. All three agent integrations share it so they cannot drift.
- **`scripts/agents/invoke-agent-worktree-guard.ps1`** — thin Claude/Codex/Copilot `PreToolUse` adapter. Normalizes payloads and responses only.
- **`.claude/hooks/pre-bash-guard.sh`** — fixed to read stdin (it read `CLAUDE_TOOL_INPUT`, so the destructive guard was inert on the native path). Retained as a fast candidate-token shim: noncandidate commands never start PowerShell.
- **`.codex/hooks.json`** — Codex `PreToolUse`, Bash matcher only; `commandWindows` runs one explicit PowerShell process.
- **`.github/hooks/hooks.json`** — Copilot entry gains a `powershell` field alongside `bash`.
- **`.githooks/pre-commit{,.ps1}`** — narrow agent-scoped backstop, active only when an agent session marker is present.
- **Docs** — `docs/agents/cross-agent-git-guardrails.md`, `AGENTS.md` Git Workflow rule, `.agents/worktrees/SKILL.md`.

Git mutations are allowed only from a linked worktree that is a direct child of `.claude/worktrees` or `.worktrees` **and** carries a valid `scripts/.env.worktree` manifest. Explicit safety denials run before the location bypass.

## Review round (last commit)

A local review found practical bypasses; all are fixed with regression tests:

- **Safety rules were regexes over the raw command**, so anything between `git` and the subcommand skipped them — `git -C . reset --hard`, `git.exe reset --hard`, `git --no-pager checkout .`, `git -C . clean -fd`, `git -C . push origin -f`. `AHKFLOW_ALLOW_MAIN=1` then downgraded the location denial to a warning, breaking the documented location-only guarantee. Now classified from parsed tokens.
- **Location was decided against the payload's `cwd`**, so `cd <main> && git commit` was allowed from a worktree. The command is now segmented and `cd`/`chdir`/`pushd`/`Set-Location` tracked in order; leading `NAME=value` prefixes no longer hide the `git` token.
- **Copilot's guard never ran on Windows.** Copilot selects `powershell` on Windows and skips an entry with neither field, so the `bash`-only registration was dead on the platform this repo is developed on.
- **`core.hooksPath` was installed relative.** Git resolves `.githooks` inside whichever linked worktree runs the hook; setup now installs the main checkout's absolute path and upgrades an existing relative value.
- **Read-only inspection was denied** — `git branch --list`/`--contains`, `git tag -v`, `git config get`.
- Found while fixing: `'…' -contains $char` tests the whole string, so both backslash-escape branches in the tokenizer were dead code.

### Latency budget revised

The plan applied one 650 ms figure to two different paths. Measured warm p50 (Windows 11, PowerShell 7): noncandidate 54 ms, shim read-only Git 725 ms, shim mutating Git 975 ms, **Codex direct PowerShell mutating 630 ms**. 650 ms was written as the Codex startup budget and Codex meets it; the Bash-shim path cannot, because the floor is ~260 ms of `pwsh` startup plus ~200–340 ms more when Git Bash spawns it. Closing that gap needs either policy duplicated into Bash (drift) or PowerShell on every command (54 ms → ~500 ms common case) — both rejected by the design. Budget is now split 650 ms Codex / 1000 ms shim, with the measurements recorded in the plan and docs.

## Verification

- `tests/AgentWorktreeGuard.Tests.ps1` (172 cases) and `tests/AgentPreCommitHook.Tests.ps1` — pass under PowerShell 7 and Windows PowerShell 5.1. Both added to the CI Windows job.
- Existing worktree, pre-push, and skill-parity suites — pass.
- Release build 0 warnings; `dotnet test` green; `dotnet format AHKFlowApp.slnx --verify-no-changes` clean.
- Every mutation test runs in a disposable repository under the system temp directory; none touch the real checkout.

## Known gaps (deliberate)

- `git commit --no-verify`, a replaced `core.hooksPath`, shell aliases, and wrappers such as `pwsh -File custom.ps1` remain bypasses. This is accidental-misuse protection, not hostile-process isolation.
- Adapter parse errors and unexpected location-policy errors fail open with a warning; only an explicit safety match fails closed.
- Agents may still edit, build, test, and format in main — the guard blocks Git mutation, not a dirty tree.
- `AHKFLOW_GUARD_DISABLE=1` disables the whole command guard by design, for emergency recovery.

## Not yet verified

The real-session Copilot adapter probe (Task 2 Step 10) could not run — organization policy rejects the session. The Windows path is currently covered by config-shape assertions only.

## Follow-up

`feature/wt-agent-worktree-only-enforcement` and its worktree are intentionally retained until this is judged to supersede them; cleanup is a separate decision.


s205109:
## Review round 2 — 4 findings, all accepted and fixed in `935cd6e`

A local review found four more gaps. All four reproduced; none were false positives. No pushback this round.

### P1 — Quoted/path-qualified Git bypassed the Bash shim

`.claude/hooks/pre-bash-guard.sh` used an enumerated delimiter class (`[[:space:];&|()\`]`) as the leading boundary. Quotes and path separators are not delimiters, so these exited 0 in Bash without reaching policy:

```
"git" commit -m test
'git' commit -m test
"C:\Program Files\Git\cmd\git.exe" commit -m test
```

Reproduction found two more the review didn't list — `/usr/bin/git commit` and `/c/Program\ Files/Git/cmd/git.exe commit`.

The important detail: **the PowerShell core already denied all of them correctly.** Only the fast-path prefilter was wrong — the worst place for it, since it fails open and silently. The file's own comment states the intended tradeoff ("a false positive costs one jq call; a false negative would silently disable the guard") but the pattern didn't honour it. Both boundaries are now "not an identifier character".

### P1 — Failed `cd` and unhandled `popd` approved mutations that run in main

Two defects in the same state model:

- A `cd` to a nonexistent path **fails** and leaves the shell where it was. The guard assumed every parsed `cd` succeeded, so `cd C:\missing; git commit` from main was classified against the harmless outside path and allowed — while the commit actually landed in main.
- `popd` was an unrecognized command, so `pushd main; pushd managed; popd; git commit` left the guard believing the shell was still in the managed worktree.

Now: the effective directory only moves when the target exists as a directory, and `pushd`/`popd` are tracked as a stack. `popd` on an empty stack leaves the directory unchanged, matching the shell.

### P1 — Force-push missed clustered flags and `+` refspecs

`git push -fu`, `git push -uf`, and `git push origin +main:main` all returned safety `Allow`, so `AHKFLOW_ALLOW_MAIN=1` degraded them to a location *warning* — the same location-only-override violation this branch fixed for `reset --hard` in round 1, in a spelling that survived. Now matches clustered short flags and a leading `+` on a positional refspec.

Note this is a deliberate **broadening** past the original hook's behaviour: the old regex didn't catch `+refspec` either. Force-push protection is the intent, so the broadening is the point.

### P2 — POSIX setup still installed a relative hook path

Round 1 fixed `setup-cross-agent-skills.ps1` but missed `setup-cross-agent-skills.sh`, which still ran `git config core.hooksPath .githooks`. Demonstrated in a scratch repo — from a linked worktree the relative value resolves to `<worktree>/.githooks`, not main's, so linked worktrees would execute their own stale hooks. Now derives the main checkout from `--git-common-dir` and installs the absolute path, mirroring the PowerShell script.

## Verification

- Guard suite **196 cases**, pass under PowerShell 7 and Windows PowerShell 5.1 (was 173).
- Pre-commit suite pass under both hosts.
- 7 existing PowerShell suites pass.
- `bash -n` on the POSIX script; main-checkout derivation verified in a scratch repo with a linked worktree.
- Release build 0 warnings; `dotnet format AHKFlowApp.slnx --verify-no-changes` clean; `git diff --check` clean.

## Still unverified

The real-session Copilot adapter probe (plan Task 2 Step 10) remains blocked by organization policy. The Windows Copilot path is covered by config-shape assertions only — unchanged from the PR description, and worth weighing before merge.

PR #235, opened 2026-07-30
fix: scope agent Git guard to collision risk, add in-session prompts

s205109:
## Summary

Relaxes the agent Git guard so it gates on **collision risk**, not on "is this authoring?".

The guard previously blocked every mutating Git command in the human-owned main
checkout unless the agent was inside a managed linked worktree. That scope was drawn
against the wrong threat and was measurably too wide — it denied `git worktree
remove -h`, a help flag that changes nothing, and a routine `/clean_gone` run was
denied at `git branch -D <gone-branch>` with no way forward inside the session.

The new test: **can this command change the human's HEAD, index, or working tree in
main?** Three tiers:

- **Tier 2 — always allowed, even from main, no prompt.** `worktree prune`;
  `worktree add`/`remove` without `--force`/`-f` (and, for `add`, without `-B`);
  `branch -d`/`--delete` without any force spelling; a plain `branch <name>` create
  with zero flags; `remote prune <name>`.
- **Tier 1a — in-session permission prompt** instead of a silent denial. Everything
  else currently guarded. `push` stays here deliberately: TEST auto-deploys on push
  to `main`, a separate risk axis from the collision test.
- **Tier 1b — hard denial, never a prompt.** `git commit` only. A `PreToolUse`
  approval cannot reach the `.githooks/pre-commit` backstop that runs after, so
  offering a prompt would produce a worse failure than today's: the user approves,
  and the commit dies anyway at the backstop.

Destructive-command safety rules (force push, hard reset, forced clean, discarding
`checkout .`, recursive-force file deletion) are untouched and still evaluate first.
`AHKFLOW_ALLOW_MAIN=1` keeps its exact previous meaning for every tier including
`commit`.

Adapter behaviour differs by necessity: Claude and Copilot emit
`permissionDecision: "ask"`; **Codex must not** — `ask` there is "parsed but not
supported yet" and makes the hook run fail *open*, letting the mutation through, so
every Ask-tier decision renders as `deny` on Codex. Subagent calls also never get a
prompt (they hard-deny and report back for a main-thread retry), and any unrecognized
decision action fails closed on all three adapters.

Also fixes the docs that asserted the old policy — including one instruction that
could never work: `.agents/worktrees/SKILL.md` told agents to recover with an inline
`AHKFLOW_ALLOW_MAIN=1 git commit -m "..."` prefix, which the guard never sees, because
it runs in its own process and reads the command as text.

## Checklist

- [x] [Canonical pre-PR gate](../docs/development/testing-workflow.md#canonical-pre-pr-gate) passes (build, format, coverage, `git diff --check`)
- [x] The change was exercised, not just compiled — name the test or verification below
- [x] No new warnings introduced
- [x] Breaking changes documented (if any) — none; the change only widens what is permitted

## Verification

Gate, run on `07541fbd`:

- `dotnet build AHKFlowApp.slnx --configuration Release` — succeeded, 0 warnings
- `pwsh .\scripts\test-fast.ps1 -Mode Coverage` — all per-assembly thresholds met (line 94.1%, branch 80.7%)

Guard behaviour (the surface this PR actually changes):

- `tests/AgentWorktreeGuard.Tests.ps1` — **301/301**. New coverage pins every tier
  boundary: Tier 2 allows; each force variant returning Ask rather than Allow
  (`-D`, `-d -f`, `-df`, `-fd`, `--delete --force`, `-m`/`-M`, `worktree add -B`,
  forced `worktree remove`, `move`/`repair`/`lock`/`unlock`); `-b` staying Tier 2 so
  the deliberate case-sensitivity of the `-B` check is pinned in both directions;
  safety rules never downgrading to Ask; the per-adapter Ask contract (Claude `ask`,
  Copilot `ask`, Codex `deny`); the subagent Ask-to-Deny downgrade with a
  no-`agent_id` control; and commit dominance across chained commands (a staging
  command followed by `git commit` denies as a whole, in both orderings and with a
  semicolon separator).
- `tests/AgentPreCommitHook.Tests.ps1` — **11/11**, including a new case pinning that
  `commit` from main is denied at the `PreToolUse` layer *before* the backstop is
  reached, so nobody later converts `commit` to a prompt.
- `tests/SkillParity.Tests.ps1`, `tests/SkillLayout.Tests.ps1` — pass (25 active
  skills). `tests/CodexSkillsHashParity.Tests.ps1` self-skips on Windows by design
  (Linux-only bash comparison); it runs in CI.

**Outstanding, deliberately not claimed:** the `Ask` path has not yet been exercised
in a live agent session — all Ask evidence above is the JSON contract asserted
in-process. Tests cannot cover whether `permissionDecision: "ask"` actually renders
as a prompt. Two manual checks from the main checkout are the last gate before merge:
`git worktree prune` should run with no prompt, and `git branch -D <gone-branch>`
should raise an approval prompt. Reviewer/author to confirm before merging.

### Footnote from writing this PR

Creating this PR was itself blocked by the guard's `dangerous-rm` rule: the summary
above originally spelled out the recursive-force deletion flags in backticks, and the
tokenizer treats a backtick as a segment separator without understanding heredocs, so
it read the prose as an invocation. Consistent with the documented "the command
tokenizer is not a full Bash/PowerShell parser" limitation, and harmless here — the
body was passed via a file instead. Noted in case it recurs often enough to be worth
addressing separately.

### DIFF
--- a/.agents/worktrees/SKILL.md
+++ b/.agents/worktrees/SKILL.md
@@ -189,10 +189,23 @@ scripts/agents/setup-cross-agent-skills.ps1 to regenerate them. -->
 
 ## Git guardrails: mutate only in a managed worktree
 
-Agents may inspect, edit, build, test, and format in the main checkout, but **Git mutations**
-(branch, add, commit, merge, rebase, push, tag, reset, …) for this repository are allowed only
-from a **managed** linked worktree. This is enforced by the `PreToolUse` command hook and, after
-merge, a `pre-commit` backstop. See `docs/agents/cross-agent-git-guardrails.md`.
+Agents may inspect, edit, build, test, and format in the main checkout. Most **Git mutations**
+(branch, add, commit, merge, rebase, push, tag, reset, …) for this repository still need either a
+**managed** linked worktree, or — new — an in-session approval prompt if run from the main
+checkout. This is enforced by the `PreToolUse` command hook and, after merge, a `pre-commit`
+backstop. See `docs/agents/cross-agent-git-guardrails.md`.
+
+A short list of operations is always allowed, even from the main checkout, with no worktree and no
+prompt: `git worktree prune`; `git worktree add`/`remove` without `--force`/`-f`; `git branch
+-d`/`--delete` without any force spelling; a plain `git branch <name>` create with no flags; `git
+remote prune <name>`. Force or destructive variants of these are **not** on this list — for example
+`git worktree remove --force`, `git branch -D`, and `git branch -m`/`-M` still need the prompt
+below, like every other guarded operation.
+
+`git commit` is the one exception that always needs a worktree, or a session-wide
+`AHKFLOW_ALLOW_MAIN=1` (see below). It can never be approved through the prompt: a separate
+`.githooks/pre-commit.ps1` backstop runs after a prompt would be approved, and that backstop has no
+way to know the prompt happened — it would still block the commit.
 
 A worktree is **managed** only when all three hold:
 
@@ -202,7 +215,9 @@ A worktree is **managed** only when all three hold:
 
 ### Recovering from a denial
 
-The stderr message is `BLOCKED: agent Git mutations are allowed only in a managed linked worktree.`
+The stderr message for a hard denial is `BLOCKED: agent Git mutations are allowed only in a
+managed linked worktree.` Most other denials show up as an in-session permission prompt instead —
+read on for which is which.
 
 1. Create a proper worktree and run the command there:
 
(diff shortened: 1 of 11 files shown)


s205109:
## Manual `Ask`-path verification could not be completed — and why

The PR body lists two manual checks as the last gate before merge. Attempting them surfaced a
**pre-existing bypass unrelated to this branch**, which makes those checks unable to prove
anything in this environment as currently configured.

### 1. The checks cannot run against this branch from the main checkout

The hook resolves its script paths from the project directory, so a session rooted in the main
checkout runs *main's* copy of the guard — currently `ba519201`, i.e. the old code. Testing there
exercises the old guard, not this branch. (A first attempt at `git worktree prune` from main ran
silently, which the old code should have hard-denied — the first hint something was off.)

### 2. `rtk git …` defeats the guard's git classification

This environment registers a user-level `PreToolUse` Bash hook, `rtk hook claude`, which rewrites
`git …` into `rtk git …`. The guard identifies a git invocation from each command segment's
leading word, so that word becomes `rtk` and no git invocation is found.

Measured directly against the policy core on this branch, `Cwd` = the main checkout:

| Command | Action | Rule |
|---|---|---|
| `git worktree repair` | Ask | `agent-main-git-mutation` |
| `rtk git worktree repair` | **Allow** | `none` |
| `git commit -m x` | Deny | `agent-main-git-mutation` |
| `rtk git commit -m x` | **Allow** | `none` |
| `git branch -D topic` | Ask | `agent-main-git-mutation` |
| `rtk git branch -D topic` | **Allow** | `none` |

So the Tier 1b hard denial on `commit` — the invariant this branch is most careful about — does not
hold for the wrapped form. This affects the old guard identically; it is not a regression
introduced here. It is the documented "wrappers … remain bypasses" limitation in
`docs/agents/cross-agent-git-guardrails.md`, except that in this environment the wrapper is the
default path for **every** git command rather than an unusual case.

Related: a `gh pr create` invocation *was* correctly caught by the `dangerous-rm` safety rule,
because only the first word of a command receives the `rtk` prefix — an `rm` appearing in a later
segment is still classified. The bypass is specific to the wrapped command's own leading word.

### 3. One observation left unexplained

A path-qualified probe (`"C:/Program Files/Git/cmd/git.exe" -C <main> worktree repair`), which the
`rtk` matcher should not rewrite, also completed without a prompt or a denial — even though both
the old copy (Deny) and this branch's copy (Ask) classify it as blocking when evaluated in
process. Why the decision did not reach the harness is not established, and is not guessed at here.

### Suggested handling

- **Do not block this PR on the manual checks.** The in-process contract is asserted by 301 passing
  cases, including the per-adapter `Ask` payload shape. What cannot be confirmed is the last mile —
  that Claude renders `permissionDecision: "ask"` as a visible prompt.
- **Treat the `rtk` interaction as separate follow-up work.** It predates this branch, applies to
  the old guard equally, and needs its own decision: either teach the guard to look through a
  known wrapper prefix, or exclude `git` from the `rtk` rewrite. Both are design choices beyond
  this PR's scope.
- Worth adding to the accepted-limitations list that a command-rewriting wrapper hook registered
  outside the repository can silently neutralise the guard, since that is materially different from
  a user deliberately invoking a wrapper script.

PR #237, opened 2026-07-30
fix: surface Blazor boot failures instead of freezing at 99%

s205109:
## What

Blazor is now started by hand from `wwwroot/js/bootBlazor.js` instead of autostarting. That script owns the boot outcome:

- It logs the original error object with `console.error`, which autostart swallowed into an uncaught promise rejection.
- It reloads the page once, guarded by the `sessionStorage` key `ahkflowapp-boot-retry-reload`.
- On a second failure in the same tab it replaces the frozen loading circle with a plain-HTML error screen.
- A successful boot clears the guard key, so a later unrelated failure keeps its free retry.

No C# changed.

## Why

A failed asset download left the loading circle frozen at 99% with nothing in the UI and nothing in the console. The boot UI had no failure branch at all.

## Two changes against the plan

Both were forced by what the browser actually did, and both are backed by test runs.

1. **`Blazor.start()` never rejects on a failed asset download.** With the app assembly 404ing, the .NET runtime reports the failure as uncaught errors (`Failed to start platform. Reason: ...`) at 0.3s, and the promise still had not settled after 120s. A `.catch` alone can never see this failure. So the script also treats an uncaught `error` or `unhandledrejection` before boot completes as a boot failure. It waits 5 seconds first, so a stray error in a boot that still succeeds does not trigger the screen.
2. **Playwright route globs did not match.** `**/_framework/AHKFlowApp.UI.Blazor.*.wasm` never fired against the absolute request URL in Playwright 1.59, so the fault was never injected. `BootFault` now uses a regular expression, which does fire.

## Tests

`tests/AHKFlowApp.E2E.Tests/BootFailureFlowTests.cs` covers four cases against a real 404 on the app assembly, injected with Playwright route interception:

- A successful boot still renders, and clears the guard key.
- The first failure reloads the page.
- The second failure shows the message, stops reloading, and the Reload button requests the document again.
- With `sessionStorage` blocked, the message appears with no reload.

The error screen copy is asserted word for word, so a reworded screen fails the test.

### DIFF
--- a/docs/environments.md
+++ b/docs/environments.md
@@ -52,3 +52,4 @@ Server=tcp:{SQL_SERVER_FQDN},1433;Database={SQL_DATABASE_NAME};Authentication=Ac
 - **TEST/PROD: deploy fails** — verify GitHub secrets/variables have `_TEST` / `_PROD` suffix, the deployer UAMI has RBAC, and the SQL firewall allows the runner IP.
 - **TEST/PROD: API returns 500** — `az webapp log tail --name <APP_SERVICE_NAME> --resource-group <RESOURCE_GROUP>` and check Application Insights.
 - **Wrong appsettings loaded** — `ASPNETCORE_ENVIRONMENT` is case-sensitive (`Development` / `Test` / `Production`).
+- **Frontend shows "Couldn't load the app"** — one or more `_framework` files failed to download. The page already reloaded itself once. Reload again; if that does not help, check the browser console for the failing request.
--- a/src/Frontend/AHKFlowApp.UI.Blazor/wwwroot/index.html
+++ b/src/Frontend/AHKFlowApp.UI.Blazor/wwwroot/index.html
@@ -32,10 +32,11 @@
         <span class="dismiss">🗙</span>
     </div>
     <script src="_content/Microsoft.Authentication.WebAssembly.Msal/AuthenticationService.js"></script>
-    <script src="_framework/blazor.webassembly.js"></script>
+    <script src="_framework/blazor.webassembly.js" autostart="false"></script>
     <script src="_content/MudBlazor/MudBlazor.min.js"></script>
     <script src="js/downloads.js"></script>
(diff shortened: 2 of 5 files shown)


s205109:
Review finding addressed in `df782082`.

**Finding (P2, confirmed):** the `error` / `unhandledrejection` path treated *any* uncaught error during a pending boot as a boot failure, so a boot slower than the 5-second grace plus an unrelated error reloaded the page and then showed a wrong "files could not be downloaded" message.

**What the runtime actually does**, measured across twelve loads with the app assembly returning 404:

- The `Blazor.start()` promise never settles — still pending after 120 seconds.
- The failure arrives as `unhandledrejection`.
- Every failing load produced `download '<url>' for <name> failed …` rejections, repeated, because the runtime retries the download.
- Only *some* loads also produced the terminal `Failed to start platform. Reason: …`. Matching that message alone was unreliable and made the suite flaky (1 in 3 runs failed).

**Fix:** only those two runtime messages count as a boot failure. `Failed to start platform` is terminal and is reported at once. A download failure waits 10 seconds and is reported only if the boot still has not finished, because the runtime may still recover. An unrelated uncaught error is ignored, however long the boot takes.

**Regression test:** `SlowBoot_WithUnrelatedUncaughtError_StillRendersApp` — the app assembly is delayed 8 seconds and an unrelated error is thrown at 500 ms. Verified red before the fix (`Expected documents.Count to be 1, but found 2`), green after.

**Correction to the PR description above:** its claim that a stray error is ignored "so a stray error in a boot that still succeeds does not trigger the screen" was only true for boots finishing inside the grace period. That is now true regardless of boot duration.

**Gate:** build 0 warnings/0 errors, `dotnet format --verify-no-changes` clean, `node --check` clean, Coverage slice 3214 tests green, `BootFailureFlowTests` 5/5 green on five consecutive runs.

The plan and spec were updated to match the shipped design (private plans repo, commit `fa29426`).

PR #257, opened 2026-08-04
chore: vendor and adapt mp-prototype and mp-research skills

s205109:
## What

Vendors two skills from `mattpocock/skills` — `mp-prototype` and `mp-research` — and adapts them
to this repo. Pinned at upstream commit `9603c1cc8118d08bc1b3bf34cf714f62178dea3b`, the same
baseline as the seven `mp-*` skills already here.

Skills only. No compiled code changes.

## Why the adaptation goes deeper than the other seven

The existing seven `mp-*` skills needed only mechanical changes: cross-skill reference rewrites and
description trimming. These two are different. Upstream writes them around a JavaScript stack, so
they told an agent in this repo to call APIs that do not exist here:

| Upstream said | This repo has |
|---|---|
| `searchParams.get('variant')` in a `tsx` block | `[SupplyParameterFromQuery]` |
| `router.replace` on Next, `navigate` on React Router | `NavigationManager.GetUriWithQueryParameter` |
| `process.env.NODE_ENV !== 'production'` | the `DEBUG` compilation symbol |
| `pnpm run <name>`, `package.json` scripts | `dotnet run --project <path>` |

Left as-is, an agent following `mp-prototype` would write code that cannot compile. Upstream prose
is otherwise untouched, so a future selective merge still lines up. `.agents/ATTRIBUTION.md` records
exactly what diverged and why.

`mp-research` gains an "In this repo" section instead. It covers the facts a research agent here
would otherwise rediscover the hard way:

- `autohotkey.com` 403s automated fetchers — fetch from `AutoHotkey/AutoHotkeyDocs` via `gh api`
- `docs/development/ahk-v2-syntax.md` already documents the AHK subset we emit; read it first
- pin API versions from `Directory.Packages.props` before citing external docs
- the main agent reviews citations, not the background agent

## The "No tests" question

`mp-prototype` rule 4 says "No tests." That reads as a conflict with **Verification After
Implementation** in `AGENTS.md`. It is not, and the skill now says so: that section governs code
that ships, and prototype code never ships. The durable test arrives with the rewrite in rule 6,
when the validated decision is folded into real code. The rule stayed; the boundary is now explicit
rather than something a reader has to infer.

## Verification

`AGENTS.md` exemption 1 applies — skills and docs only, nothing compiled.

- `SkillLayout.Tests.ps1` and `SkillParity.Tests.ps1` pass; 27 active skills
- Working-tree parity between `.agents/` and the plugin projection verified with `git hash-object`
  on every edited file
- Codex plugin hash rebumped to `0.1.0+codex.fbe0b4c873f1`
- Descriptions are 127 and 101 chars, inside the 140-char budget at
  `scripts/agents/setup-cross-agent-skills.ps1:182`
- Blazor APIs checked against Microsoft Learn for `aspnetcore-10.0`, not from memory

## For the reviewer

Worth a close look:

1. Is the deeper adaptation the right call, or should these two have stayed closer to upstream? The
   fork policy in `ATTRIBUTION.md` is the thing being stretched.
2. `mp-research`'s "In this repo" section duplicates facts that also live in `AGENTS.md`. Useful
   where the agent reads it, or a second copy that will drift?

## After merge

Run `pwsh .\scripts\agents\setup-cross-agent-skills.ps1` from the main checkout, then restart Codex.
The plugin cache currently tracks this worktree.


s205109:
## Review round 2 — all six findings fixed

Commit `987fde2d`. Every finding verified against the code before editing; all six confirmed.

| # | Verdict | Fix |
|---|---|---|
| 1 | CONFIRMED | `UI.md:68-81` — renamed to `PrototypeVariant`; collision explained at `:87-91` |
| 2 | CONFIRMED | `mp-research/SKILL.md:26` — `?ref=v2` added; reason at `:29` |
| 3 | CONFIRMED | `mp-research/SKILL.md:21` — internal doc is orientation, primary source is the citation |
| 4 | CONFIRMED | `UI.md:66-77` — whole subtree gated, real page as production fallback; `UI.md:122` — `IsDevelopment()` replaces `DEBUG` |
| 5 | CONFIRMED | `mp-prototype/SKILL.md:29-37` — invented exemption removed, build-and-exercise now required |
| 6 | CONFIRMED | `SKILL.md:39`, `UI.md:132` — capture first, fold second |

### One correction to finding 2

`-f ref=v2` does not work. `-f` makes `gh` send a POST, and the contents endpoint answers 404:

```
$ gh api repos/AutoHotkey/AutoHotkeyDocs/contents/docs/Hotstrings.htm -f ref=v2
gh: Not Found (HTTP 404)
```

`?ref=v2` returns `d2bc3e43…` exactly as you said, and `-X GET -f ref=v2` also works. The skill now
gives the query-string form and explains why the bare `-f` fails, so nobody reaches for it again.

### A sibling defect finding 1 did not name

The same example injected `IWebAssemblyHostEnvironment` with no
`@using Microsoft.AspNetCore.Components.WebAssembly.Hosting`. That namespace is absent from
`_Imports.razor`, which is why `Pages/Settings.razor:3` declares it per file. Same "does not compile"
class as the `Variant` shadowing. Fixed at `UI.md:63`, explained at `:93-95`.

Checking that also sharpened finding 1 itself: `_Imports.razor` does carry `@using MudBlazor`, so the
`Variant` collision threatens every component in the app, not just Settings. The warning says so now.

### Verification

- The corrected `gh` command was run. It returns
  `<title>Hotstrings - Definition &amp; Usage | AutoHotkey v2</title>` — v2 content, as intended.
- `SkillLayout.Tests.ps1` and `SkillParity.Tests.ps1` pass; 27 active skills.
- Working-tree parity between `.agents/` and the plugin projection confirmed with `git hash-object`
  on every edited file.
- `git diff --check` clean. Plugin hash rebumped to `0.1.0+codex.6f0a9916f18c`.

### Unrelated, pre-existing

`dck-verify` (174), `playwright-cli` (162) and `impeccable` (897) are over the 140-char description
budget. The `[WARN]` in the setup output belongs to them, not to this branch. Happy to take it as a
separate item.

PR #266, opened 2026-08-06
feat: warn when a profile template already uses a hotkey's keys

s205109:
## What

Warns an owner when a Hotkey row uses keys that one of its Profiles already uses in a header or footer template.

The shipped `capslock-modifier-layer` header preset writes `*CapsLock::` straight into the header. A row on the same key then stops working as written, and the app said nothing. Closes backlog 058.

## How

Two pure helpers in the Blazor project:

- `TemplateKeyUses.Parse` reads a template into `TemplateHotkey(Key, Wildcard, Ctrl, Alt, Shift, Win)`. It skips comments, hotstrings, and custom combinations.
- `TemplateUseText.TextFor` composes the sentence. It names at most three Profiles and counts the rest.

`ShortcutWarning.razor` canonicalizes both sides and renders a third notice inside its existing alert. Both call sites — the edit dialog and the grid's inline row — already hold the full `ProfileDto` list, so nothing new is fetched and no server code changes.

Matching follows AutoHotkey. A wildcard template hotkey fires while extra modifiers are held, so it matches a row carrying at least its modifiers. Every other template hotkey matches only the same modifiers, because `^!c::` is exactly what the emitter writes for a Ctrl+Alt+C row.

The notice never blocks a save, and it never promises an outcome. It closes with `Your hotkey may not fire.`

## Review round

The first implementation matched on the key alone. Code review found that produced both false positives and false negatives, so matching now reads the modifiers too. The spec's decisions 6 and 8 were rewritten to say why.

Also in this round: the overflow subject no longer folds an unnamed footer hit under a header subject, and the throwaway prototype is deleted.

One finding was deferred. `LControl` and `RControl` are valid AutoHotkey names that the alias map does not accept. Adding them breaks `HotkeyTypedActionsMigrationTests`, because the fixtures come from `HotkeyKeys.Aliases` while the applied migration carries a frozen name list. That needs its own decision — recorded as `backlog/059-lcontrol-rcontrol-key-aliases-missing.md`.

## Verification

Canonical pre-PR gate, all four steps:

- `dotnet build AHKFlowApp.slnx --configuration Release` — 17 projects, 0 warnings, 0 errors
- `pwsh .\scripts\test-fast.ps1 -Mode Coverage` — 3582 passed, 0 failed; line 94.6%, branch 82.4%

New coverage: 12 parser tests, 16 composer tests, 8 component tests across the dialog and the inline row, and one E2E flow that edits a profile header, binds the same key, reads the notice, and still saves.


s205109:
## Review response — @opencode-agent

Thanks for the detailed pass. Verdicts below, one per point. Fix commit: `63bf433c`.

| # | Finding | Verdict | Result |
|---|---|---|---|
| 1a | `*CapsLock:: ; comment` dropped | Pushback | Already parsed |
| 1b | `CapsLock::{` dropped | Pushback | Already parsed |
| 1c | `^~c::` dropped (prefix after modifier) | Confirmed | Fixed |
| 2 | `#HotIf` not modeled | Partial | Documented, not modeled |
| 3 | `Profiles` not in `EvaluationKey` | Pushback | Latent only; assumption documented |
| 4 | Branch name / test count | Partial | Count corrected below; branch left alone |

### 1a and 1b — already handled

`HotkeyOn` splits the line at `::` and reads only the left side (`TemplateKeyUses.cs:74`, `string left = line[..doubleColon];`). A trailing comment and a same-line `{` both sit to the **right** of `::`, so neither reaches the `rest` check at `TemplateKeyUses.cs:132`.

Both are now covered by tests, so the claim stays disproved:

- `TemplateKeyUsesTests.Parse_ReadsAHotkeyWithATrailingComment`
- `TemplateKeyUsesTests.Parse_ReadsAHotkeyWithAnOpeningBraceOnTheSameLine`

Both passed before the fix commit and still pass.

### 1c — confirmed and fixed

This one was real. The prefix loop ran to completion before the modifier loop started, so `^~c::` hit `~` in the modifier switch, jumped to `done`, and found no key character. Line dropped.

The two loops are now one pass over prefixes and modifiers together, which is how AutoHotkey reads them: `TemplateKeyUses.cs:95` (`case char p when Prefixes.Contains(p):`).

Regression tests, red before the fix and green after:

- `TemplateKeyUsesTests.Parse_ReadsAPrefixThatFollowsAModifier` (`^~c::`)
- `TemplateKeyUsesTests.Parse_ReadsAWildcardThatFollowsAModifier` (`#*e::`)

### 2 — `#HotIf` stays unmodeled, now documented

Agreed that this can produce a false positive, and agreed it is acceptable for an advisory notice. Modeling window context would mean tracking `#HotIf` scope through the template and then comparing it against a row's own window context — a larger change than this branch should carry.

It is now a stated limit instead of a silent one:

- `TemplateKeyUses.cs:26-40` — a "Known limits" block on the class, covering both the `#HotIf` false positive and the skipped-line false negative.
- `docs/development/ahk-v2-syntax.md:348-356` — the same two limits, plus the new "prefixes and modifiers in any order" rule.

### 3 — latent only, so the assumption is documented rather than coded around

The `Profiles` list is loaded once, at `Pages/Hotkeys.razor:366`, and that is its only assignment in the file. The same list instance goes to the inline row (`Pages/Hotkeys.razor:829`) and to the dialog (`Pages/Hotkeys.razor:698` and `:730`). Nothing replaces it while a notice is on screen.

Putting template content into `EvaluationKey` would mean hashing every selected profile's header and footer on every parameter set, for a path that no call site reaches. Instead the contract now says the list is fixed for the page's lifetime: `Components/Hotkeys/ShortcutWarning.razor:65-70`. If a future call site reloads profiles in place, that comment is where it will be caught.

### 4 — count corrected, branch left alone

The test count is off in both the PR body and the review. The branch adds **10** component tests: 9 in `HotkeyEditDialogTests` and 1 in `HotkeysPageTests`. The PR body says 8; the review says 9.

The branch name stays `fix/wt-042-…`. Renaming the head branch of an open PR is not worth closing and reopening PR 266 over a cosmetic mismatch.

### Gate

Full canonical pre-PR gate, all four steps green:

- `dotnet format AHKFlowApp.slnx --verify-no-changes` — exit 0
- `pwsh .\scripts\test-fast.ps1 -Mode Coverage` — every per-assembly threshold met; line 94.6%, branch 82.4%
- `git diff --check main...HEAD` — exit 0

PR #269, opened 2026-08-06
fix: report a failed file save on the downloads page

s205109:
## Summary
- Downloads page left a row/button disabled with no message when `IFileSaver.SaveAsync` threw a `JSException` — the busy state cleared internally, but Blazor skips its post-event render when the event task ends faulted, so the button stayed drawn as disabled.
- Same catch shape `Profiles.razor` already uses, applied to both `DownloadProfileAsync` and `DownloadAllAsync` (the zip button had the same bug, out of scope on the original backlog item but fixed here too).

Closes backlog 057.

## Test plan
- [x] `pwsh .\scripts\test-fast.ps1 -Mode Fast` — 942/942 UI.Blazor.Tests pass, including two new bUnit tests
- [x] `pwsh .\scripts\test-fast.ps1 -Mode E2E` — 53/53 pass, including two new flow tests driving the real interop path
- [x] Canonical pre-PR gate: build, format --verify-no-changes, full coverage gate, `git diff --check` — all green

### DIFF
--- a/backlog/done/057-downloads-page-row-stays-disabled-when-the-file-save-fails.md
+++ b/backlog/done/057-downloads-page-row-stays-disabled-when-the-file-save-fails.md
@@ -24,10 +24,13 @@ purpose, because backlog 055 put changes to that page out of scope. The two page
 
 ## Acceptance criteria
 
-- [ ] A failed file save on the Downloads page shows an error message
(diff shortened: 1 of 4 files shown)


s205109:
@codex[agent] review this PR

PR #275, opened 2026-08-07
fix: show hover text on disabled shortcut action buttons

s205109:
## What

Show the hover text on a known-shortcut action button while the button is disabled.

Closes backlog 053.

## Why

The three action buttons on a known-shortcut row (delete, restore, ignore) get disabled while a
write is in flight. Their hover text lived in a `title` attribute on the button. MudBlazor sets
`pointer-events: none` on a disabled button, so the browser never fired the hover and the text
never appeared — exactly when a user is most likely to hover and ask "why can I not click this?"

## How

Each `MudIconButton` is now wrapped in a `MudTooltip`. The tooltip root is a plain `<div>` that is
never disabled, so it still receives the hover. Hit testing falls through the disabled button to
that wrapper. The short text moved from the button's `title` to `MudTooltip Text`.

`aria-label`, `data-shortcut-id` and `data-used-by` stay on the button, not on the wrapper. A
`<div>` has no role, so a label on the wrapper would be ignored by a screen reader.

## Verification

- bUnit theory test, one case per action button, rendered with `Busy="true"`: asserts the wrapper
  exists, the button really is disabled, and the text appears after a real `PointerEnter`
- Playwright E2E test hovers the same wrapper in a real browser while an ignore-write is held open
  with a routed `TaskCompletionSource`
- Canonical pre-PR gate: build (17 projects, 0 errors), `dotnet format --verify-no-changes`,
  coverage (3,616 tests, thresholds met), `git diff --check main...HEAD`
- E2E slice: 54/54

## Note for reviewers

Two details in the E2E test look odd but are load-bearing:

1. `Mouse.MoveAsync(0, 0)` before the hover. `ClickAsync` leaves the pointer inside the tooltip
   root. MudTooltip closes on click and the pointer never leaves, so a following `HoverAsync` moves
   to a point it already occupies and Chromium fires no `pointerenter`.
2. The assertion targets `.mud-popover-open.use-action-tooltip-text`, not the bare class. MudBlazor
   keeps every popover mounted off-screen, so the bare class matches every row on the page.

The plan document still carries the original, non-working test code for both points. Correcting it
is follow-up work in the plans repo.

### DIFF
--- a/backlog/done/053-title-tooltip-dead-on-disabled-action-buttons.md
+++ b/backlog/done/053-title-tooltip-dead-on-disabled-action-buttons.md
@@ -27,15 +27,25 @@ name.
 
 ## Acceptance criteria
 
-- [ ] Hovering a disabled known-shortcut action button shows its short text
-- [ ] The fix wraps the button rather than styling the disabled state. `LoginDisplay.razor:17-19`
+- [x] Hovering a disabled known-shortcut action button shows its short text
+- [x] The fix wraps the button rather than styling the disabled state. `LoginDisplay.razor:17-19`
       is the pattern already in the repo — a disabled `MudButton` inside a `MudTooltip`. The
       tooltip root is a plain div, it is not disabled, so it receives the hover
-- [ ] The `aria-label` at `:53` keeps naming the use in full, so the desktop table and the mobile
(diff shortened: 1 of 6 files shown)


s205109:
@oc review this PR

s205109:
@codex review this PR

PR #278, opened 2026-08-08
fix: guard Edit/Write into main from worktree sessions (065)

s205109:
Closes backlog 065. A Claude session whose working directory is a managed worktree can no longer
`Edit`, `Write`, or `NotebookEdit` a file that resolves under the human-owned main checkout.

## The gap

Two guards decided who may write into main, and they disagreed.

| Guard | Decides from | Covered |
|---|---|---|
| Claude Code native isolation | a session flag set by `-w`, `--worktree`, or `EnterWorktree` | `Edit`, `Write` |
| This repository (backlog 054) | the `cwd` in the hook payload | `Bash` |

A session started **directly inside a worktree directory**, without `-w`, never gets the native
flag. It could `Edit` and `Write` into main freely, while the same write through `Bash` was denied.
The whole gap was one line — `invoke-agent-worktree-guard.ps1:65` exited early for every tool that
was not a shell.

## What changed

Three layers, copied from the `Bash` write guard that backlog 054 shipped.

- **`.claude/hooks/pre-edit-guard.sh`** (new). A prefilter on the payload `cwd`. A managed worktree
  always carries the `worktrees` path segment, so a main-checkout session exits in Bash and never
  pays a PowerShell start. Same conservative-superset discipline `pre-bash-guard.sh` documents.
- **`invoke-agent-worktree-guard.ps1`**. `ConvertFrom-AgentHookInput` gains a `file-edit` tool class
  and a `TargetPath` field, read from `file_path` or from `notebook_path`. The entrypoint answers a
  file-edit call directly: stderr plus exit 2 to deny, stderr plus exit 0 to warn, silence to allow.
- **`agent-worktree-guard.common.ps1`**. One new function, `Get-AgentFileEditWriteDecision`. A tool
  call carries one literal path, so it needs no parsing. It reuses 054's existing path resolution,
  allow-list, and refusal messages rather than growing a second copy.

Registered on `Edit|Write|NotebookEdit`. Rule name stays `agent-worktree-main-write` — same rule
about the same thing.

Codex and Copilot file-edit calls stay out of scope. Their tool names and payload shapes are not
verified here, and neither agent has native isolation to disagree with. Recorded as a limitation.

## Three fail-open defects this exposed in the shared resolver

Reusing 054's path resolution meant reading it closely, and it had three ways to let a write
through. Each one threw, and every throw reaches the entrypoint's catch, which **allows** the call.
All three affected the existing `Bash` rule too, so 054 is now stricter as well.

1. **A relative symlink target was resolved against the wrong directory.** Windows resolves it
   against the directory holding the link; the resolver used the hook process's working directory.
   A link inside a worktree pointing at `..\..\..\README.md` therefore looked like a path outside
   the checkout, while `Edit` followed it into main.
2. **A link that was the last path component indexed past the end of an array.** The remainder slice
   `$parts[($i+1)..($parts.Count-1)]` asks for index `$parts.Count`. Under the entrypoint's
   `Set-StrictMode -Version 3.0` that throws; without strict mode it silently produces a descending
   range and duplicates the leaf. Measured against the repository's own tracked relative symlink at
   `.github/AGENTS.md`.
3. **A path the PowerShell host cannot parse threw instead of failing closed.** Windows PowerShell
   5.1 runs on .NET Framework, whose `Path.IsPathRooted` rejects a `"` outright; pwsh 7 accepts it.
   Rooting and `GetFullPath` are now wrapped together and report the target unresolved, which
   denies.

A fourth issue was specific to the new code: `-Literal` skipped the shell-expansion rule but still
ran `.Trim().Trim('"', "'")`. A quote and a leading space are legal Windows file name characters, so
trimming named a different file — `'\..\..\..\..\victim.txt` became the drive-rooted
`C:\victim.txt`, outside the checkout, and was allowed. Literal mode now preserves the path exactly.

## Verification

Three real sessions, all from `.claude/worktrees/065-edit-write-isolation` with **no** `-w`, on
Claude Code 2.1.225. This is the exact case the item was filed on.

**Write into the main checkout — denied.**

```
PreToolUse:Write hook error: [bash "${CLAUDE_PROJECT_DIR}/.claude/hooks/pre-edit-guard.sh"]: [agent-guard:Claude] deny [agent-worktree-main-write]
BLOCKED: this session is isolated in a worktree, so it cannot write into the main checkout at C:\Dev\segocom-github\AHKFlowApp\probe-065-writetest.log
Read from here. Write from the main checkout.
To override, a human must set AHKFLOW_ALLOW_MAIN=1 in the shell environment before starting the
agent session.
```

**Write into `docs/superpowers` — denied, with the plans wording.** The symlink resolves back into
main, and the message never sends the agent looking for a worktree copy that cannot exist.

```
BLOCKED: this session is isolated in a worktree, so it cannot write into the main checkout at C:\Dev\segocom-github\AHKFlowApp\docs\superpowers\probe-065-plans.md
docs/superpowers is a symlink to the main checkout. There is no worktree copy of it, so do not
look for one. Plan and spec writes belong in the main checkout by design.
Read the plan from here. Write and commit it from the main checkout.
```

**Write inside the worktree — succeeded.** The control. Without it the first two prove only that the
guard refuses everything.

No probe file was created. Main stayed clean and the plans repository stayed clean.

Test gate:

- `tests/AgentWorktreeGuard.Tests.ps1` — pass under pwsh 7 and under Windows PowerShell 5.1
- All 16 PowerShell suites CI runs, in a fresh clone with no worktrees — all green
- Pre-push quick checks — 2,925 tests, pass

`pwsh .\scripts\test-fast.ps1 -Mode Coverage` was not run. This branch changes no compiled file, so
it cannot move coverage.

## A CI hole found while planning

`tests/AgentWorktreeGuard.Tests.ps1` had a test failing on `main`, and CI reported the job green.
`ci.yml` runs sixteen suites in one `run:` block, so only the last script's exit code counts.

The failing test assumed the suite runs inside a worktree, which is not true in a plain checkout.
It is rewritten here against a stub entrypoint, so it no longer depends on where the suite runs. The
new entrypoint tests use the same idea and run against a copy planted in the disposable fixture.

The CI hole itself is **not** fixed here — it is filed as `backlog/066`, with the trap its fix must
avoid: the suites have no explicit `exit 0` on success, so `$LASTEXITCODE` leaks from their last
internal `git` call. `WorktreeBaseRef.Tests.ps1` prints "passed" and still leaves it at `1`.

## Also in this PR

- `backlog/065` ticked and moved to `backlog/done/`
- `backlog/066` filed
- `docs/agents/cross-agent-git-guardrails.md` updated — new hook row, the write-isolation section
  now covers both kinds of write and the symlink rule, and the stale "not covered by this repository
  at all" claim is gone

### DIFF
--- a/.claude/hooks/pre-edit-guard.sh
+++ b/.claude/hooks/pre-edit-guard.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+# Pre-Edit Guard - fast cwd prefilter for Claude Edit, Write, and NotebookEdit tool calls.
+#
+# Reads the native PreToolUse payload from stdin. A managed worktree always lives directly under
+# <main>/.claude/worktrees or <main>/.worktrees, so its path always carries the "worktrees"
+# segment. A payload whose cwd does not carry it cannot be a worktree session, so it exits here
+# without paying PowerShell startup. Everything else is forwarded to the shared policy core in
+# scripts/agents/invoke-agent-worktree-guard.ps1, which makes the real decision.
+#
+# Exit 2 = block. Exit 0 = allow.
+
+if [[ "${AHKFLOW_GUARD_DISABLE:-}" == "1" ]]; then
+  echo "WARNING: AHKFLOW_GUARD_DISABLE=1; all agent command guardrails are disabled." >&2
+  exit 0
+fi
+
+INPUT=$(cat)
+
+shopt -s nocasematch
+
+# Conservative superset, applied to the raw payload so the common case never pays for a jq
+# process at all. Same discipline as pre-bash-guard.sh: a false positive costs one PowerShell
+# start, a false negative silently disables the guard. An edited file whose own path contains
+# "worktrees" costs one wasted start in a main session, which is the accepted direction of error.
+WORKTREE_SESSION_PATTERN='worktrees'
+
+if [[ ! "$INPUT" =~ $WORKTREE_SESSION_PATTERN ]]; then
+  exit 0
+fi
+
+# Narrow the match to cwd when jq is available, so the wasted start above is avoided too.
+# Correctness must not depend on jq: a parse failure forwards.
+if command -v jq >/dev/null 2>&1; then
+  CWD=$(printf '%s' "$INPUT" | jq -er '.cwd // empty' 2>/dev/null) || CWD=""
+  if [[ -n "$CWD" ]] && [[ ! "$CWD" =~ $WORKTREE_SESSION_PATTERN ]]; then
+    exit 0
+  fi
+fi
+
+SCRIPT_DIR=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+if command -v pwsh >/dev/null 2>&1; then
+  POWERSHELL=(pwsh -NoProfile -NonInteractive)
+elif command -v powershell.exe >/dev/null 2>&1; then
+  POWERSHELL=(powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass)
+else
+  echo "WARNING: agent guard could not find PowerShell; allowing edit." >&2
+  exit 0
+fi
+
+printf '%s' "$INPUT" |
+  "${POWERSHELL[@]}" -File \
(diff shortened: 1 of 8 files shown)


s205109:
@codex review this PR

PR #285, opened 2026-08-09
test: await the click before each negative assertion

s205109:
## What

Backlog 069. Three tests in `KnownShortcutsPageTests` asserted a negative straight after a click.
If the click handler had not run yet, the assertion passed for the wrong reason.

Two of them now await the click. The third is ruled out, with the reason written down.

## Why the awaited click, not a rendered signal

The backlog item asked for a rendered state to wait on first. There is a better answer.

bUnit ships an async twin of every event helper. `ClickAsync` is documented as returning
"A task that completes when the event handler is done"
(`~/.nuget/packages/bunit/2.7.2/lib/net10.0/bunit.xml:2614-2622`). The synchronous `Click()`
at `:2606-2613` has no `<returns>` at all. That is the whole difference, and it is the bug
backlog 068 named.

Awaiting is exact rather than a proxy. It also covers the case no rendered signal can: the
cancelled-delete path renders nothing to wait on.

## The premise was slightly too strong

The item said these assertions "cannot fail". On an idle machine they can, and they did.
They were timing-dependent, not dead.

A probe pinned the real hole down. The failing API call was stubbed with a genuinely async task
(`await Task.Delay(50)`) — the condition a loaded machine produces. Two tests then ran against the
same page, broken so a server-answered failure invalidates the dialog cache:

- `OldForm_SyncClick_MissesTheBrokenPage` — **passed**. Blind.
- `NewForm_AwaitedClick_CatchesTheBrokenPage` — **failed**. Caught it.

The probe was temporary and is not in this branch.

## Mutation experiments

Each assertion was shown to fail against a page broken on purpose. Full output in the backlog file.

| Break | Test | Result |
|---|---|---|
| `if (confirmed != true)` removed from `DeleteAsync` | cancelled delete | FAIL 1/1 |
| `Invalidate()` moved outside the `NetworkError` check | ignore, restore, delete | FAIL 3/4 |
| `Invalidate()` added to the `CommitAddAsync` failure branch | create | FAIL 1/4 |
| `if (_isAuthenticated)` removed from `OnInitializedAsync` | signed out | FAIL 1/1 |

The create case needed its own mutation: create never reaches `AfterMutationAsync`.

## SignedOut_SaysSo_AndReadsNothing is ruled out

No click, and nothing is ever queued. The page reads inside `if (_isAuthenticated)` at
`KnownShortcuts.razor:215-216`, and the cascaded state is anonymous, so that branch is never
taken on any timeline. Mutation experiment 4 is what proves the assertion still has teeth.

## Verification

- `AHKFlowApp.UI.Blazor.Tests`: 950 of 950 pass
- 50 confidence runs against the final Release build: all 50 passed
- Canonical pre-PR gate: build, format, PowerShell suites (18 of 18), coverage thresholds, `git diff --check` — all green

No production code changed.


s205109:
@codex review this PR

PR #288, opened 2026-08-11
feat: development process wave 1 (backlog 071)

s205109:
Wave 1 of the development process redesign (backlog 071).

Spec: docs/superpowers/specs/2026-08-10-development-process-design-071.md (private plans repo)
Plan: docs/superpowers/plans/2026-08-10-development-process-plan-071.md (private plans repo)

Delivers workflow.md (canonical process), the HTML decision tree, the one-page
cheatsheet + PDF, the reusable design system, the AGENTS.md / .claude/CLAUDE.md
process-section reduction, stage stamps, wave items, and friction baselines.


s205109 (commented):
Second-round cold review of the complete PR range. CI is green, and the representations are structurally consistent, but I found 16 actionable process defects. Four can block work or lose the Ship closure records. Verdict: not ready to merge; details are in the review threads.

docs/development/workflow.md
```diff
@@ -0,0 +1,771 @@
...
+Ship flips the pull request to ready and closes the records before CI runs. So both
+non-success edges must undo what Ship already did.
+
+Failure converts the pull request back to draft
+(`gh pr ready <number> --undo --repo s205109/AHKFlowApp`). It returns to a pre-Review
+stage, and Review's entry condition is a draft pull request. Without the undo, that path is
+shut. Blocked keeps the ready pull request, because it resumes at Ship, and Ship needs a
+ready pull request.
+
+Both edges push whenever they make a commit. Ship's success is the only Ship transition
```
s205109: **Critical: the Ship closure commit never reaches the remote branch.** Ship changes the tracked item and deletes `PLAN-PROGRESS.md` at line 382, and the Stage writer requires a commit. GitHub merges the remote PR head, not an unpushed local commit. Because this rule says Ship success never pushes, the merge omits the closure records and Cleanup can delete their only copy. Push the closure commit before ready/CI/merge, or publish those records in an earlier stage.

docs/development/workflow.md
```diff
@@ -0,0 +1,771 @@
...
+
+The round pull request is created at Execute close with a `Stage: 5-simplify` line in its
+body. Exactly one such line exists from then on.
+
+```powershell
+$pr   = <round PR number>
+$next = '<N-name>'                      # e.g. 6-verify
+$rx   = '(?m)^Stage: [^\r\n]+'          # CRLF-safe: does not eat the carriage return
+
+$body = gh pr view $pr --repo s205109/AHKFlowApp --json body -q .body
```
s205109: **High: this command rejects every normal multiline PR body.** PowerShell captures multiline native `gh` output as `System.Object[]`; `[regex]::Matches($body, $rx)` then sees the array object, not one multiline string. I reproduced `type=System.Object[] matches=0`, so the command throws before `gh pr edit`. Join the lines into one scalar string before matching and replacing.

docs/development/workflow.md
```diff
@@ -0,0 +1,771 @@
...
+
+**Where Intake writes the file.** A new item written into the main checkout cannot reach a
+worktree by itself. `git worktree add` builds the new tree from a commit, and
+`scripts/new-worktree.ps1` copies only the entries listed in `.worktreeinclude`, which is
+empty. An agent also cannot commit on `main`. So an uncommitted new item in main is
+stranded.
+
+Two legal routes, and the session states which one it took:
+
+1. **Agent, tracked work — the normal route.** Create the execution location first, then run
```
s205109: **High: the normal Intake route needs the backlog number before the script assigns it.** `new-worktree.ps1` requires `-Name` before it creates the worktree, while `new-backlog-item.ps1` chooses the next number only after the worktree exists. The branch convention also requires `feature/wt-NNN-...`, and the later discovery rule requires a number-prefixed worktree. A fresh agent must guess the number or violate the naming/discovery rules. Define a race-safe reservation, rename, or filing route.

docs/development/workflow.md
```diff
@@ -0,0 +1,771 @@
...
+field in the same commit that completes the stage. A failure edge sets the field back to
+the target stage. A blocked item keeps its last stage; the move to `backlog/blocked/` plus
+the unblock note carry the rest.
+
+The field lives in the work branch during a task. The draft pull request is the live
+pointer from `main`. If branch name, pull request state, and field disagree, the field
+wins. The others are evidence.
+
+**Two-repo transitions (Design, Plan).** Commits are ordered. First the artifact to the
+private plans repo (`git -C docs/superpowers commit`), then the Stage-field update to the
```
s205109, line 679: **High: the canonical plan commit route is denied from the worktree where Design and Plan run.** `docs/superpowers/` resolves into the main checkout, and the guard refuses that write from a managed worktree (`cross-agent-git-guardrails.md:132-161`; backlog 076 records the same gap). Only `.claude/CLAUDE.md` tells Claude to return to main first, so Codex or Copilot following this canon gets blocked. State the current main-checkout handoff here until backlog 076 lands.

backlog/031-cli-winget-distribution.md
```diff
@@ -7,6 +7,7 @@
 - **Interfaces**: CLI
 - **Depends on**: 030-cli-native-windows-distribution
 - **Status**: On hold — see Notes/dependencies
+- **Stage**: 4-execute
```
s205109: **Medium: this migrated Stage value is not resumable.** Execute resume requires `PLAN-PROGRESS.md`, but the repository has no 031 worktree, local or remote branch, open PR, or progress file. Because the Stage field wins, a fresh session skips Pickup and cannot reconstruct Execute. Stamp this legacy item at Pickup, or define a migration resume rule.

backlog/042-key-rebinds-as-first-class-rows.md
```diff
@@ -6,6 +6,7 @@
 - **Type**: Feature
 - **Interfaces**: UI | API
 - **Status**: Deferred
+- **Stage**: 1-pickup
```
s205109, line 10: **Medium: this item and backlog 070 enter Pickup without `Difficulty`.** Pickup needs Difficulty to choose its location and one of the three targets. `PLAN-PROGRESS.md:11` acknowledges the fresh-session subject could not resolve that target. Add Difficulty to both migrated items, or keep them at Intake until migration supplies it.

docs/development/workflow.md
```diff
@@ -0,0 +1,771 @@
...
+```
+
+Values run from `0-intake` to `10-cleanup`. This section governs tracked items. A
+housekeeping round has no item, so it writes the same value into its round pull request
+body instead, in the same transition that would write an item's field. See section 2 for
+the read-modify-write command sequence.
+
+**Writer rule.** Whoever completes a stage transition — normally the agent — updates the
+field in the same commit that completes the stage. A failure edge sets the field back to
+the target stage. A blocked item keeps its last stage; the move to `backlog/blocked/` plus
```
s205109, line 647: **Medium: blocked items have a move into `backlog/blocked/` but no move back out.** Every resume edge describes stage work only. When the external condition clears, following resume leaves the active item in the blocked folder, so the durable location still says it is blocked. Define and commit the unblock move before stage resume.

docs/development/workflow.md
```diff
@@ -0,0 +1,771 @@
...
+- **Context** — safe to clear
+- **Review** — not reviewed
+
+| Edge | Condition | Target |
+|---|---|---|
+| success | filed, Difficulty set | 1-pickup |
+| failure | missing Difficulty or summary; complete the item | stay |
+| blocked | cannot specify without an external answer; file what is known with the unblock note | blocked/ |
+| not applicable | a housekeeping round needs no item — taken once per round, at its first change; the round's commits are the record; say so | 1-pickup |
+| resume | item sat idle; revise Difficulty once | 1-pickup |
```
s205109: **Medium: Intake resume can bypass Intake's own exit condition.** The filing script creates a placeholder summary and no Difficulty. After interruption, this edge advances to Pickup after revising Difficulty only, even though the failure row requires a missing summary to stay in Intake. Require the full Intake exit condition here, or target `stay`.

docs/development/workflow.md
```diff
@@ -0,0 +1,771 @@
...
+### When a predicate flips mid-work
+
+Stop. Re-run the full classification. The result may be `moderate`, `complex`, or
+`to-be-determined` — never an automatic `moderate`. Then move to a dedicated worktree with
+the artifacts that value needs.
+
+The reclassified change leaves its round before the round pushes. A round never grows a
+non-trivial member.
+
+- Not yet committed: move the files into the new worktree and continue there.
```
s205109: **Medium: the uncommitted reclassification route conflicts with worktree isolation.** A managed worktree session cannot move files into a sibling worktree; the guard explicitly refuses sibling writes (`cross-agent-git-guardrails.md:160-161`, proved by `AgentWorktreeGuard.Tests.ps1:1905-1908`). Use the commit/cherry-pick/drop route for both cases, or document an explicit main-session handoff.

docs/development/workflow.md
```diff
@@ -0,0 +1,771 @@
...
+A housekeeping round files no item at all, so this note does not apply to it.
+
+<a id="stage-1-pickup"></a>
+
+### Stage 1 — Pickup
+
+- **Entry** — item chosen
+- **Who** — Sonnet, default effort
+- **Technique** — `scripts/new-worktree.ps1` (`-BaseRef` for stacked work), or the housekeeping worktree
+- **Action** — choose the execution location by Difficulty: a dedicated worktree for `moderate` and `complex`, the housekeeping worktree for `trivial`. Confirm branch and base, and state both. For a tracked item: stamp Stage, push, open the draft pull request. For a housekeeping round: no item, so no Stage stamp and no push here. The round worktree is the pointer, and the route is the round pull request opened at Execute close
```
s205109: **Medium: Pickup publishes the next Stage before the PR route exists.** The action stamps, pushes, and only then opens the draft PR. If `gh pr create` fails, the authoritative remote field says Design/Plan/Execute while Pickup's `PR route in place` exit condition is false. No edge repairs that split state. Keep a durable Pickup marker until PR creation succeeds, then publish the target Stage.

docs/development/coverage.md
```diff
@@ -38,7 +38,7 @@ git config core.hooksPath .githooks
 
 The canonical gate also requires `python` on `PATH`.
 
-That `core.hooksPath` setting enables the repo-managed `.githooks/pre-push` hook, which runs `pwsh .\scripts\pre-push-quick-checks.ps1` automatically before each push — an incremental build plus the container-free fast test slice, not full coverage. CI runs the full coverage + format gate on every non-docs PR, so the hook is a quick local sanity check rather than the authoritative gate. Run `pwsh .\scripts\run-coverage.ps1` yourself before opening a PR if you want the full local confidence check.
+That `core.hooksPath` setting enables the repo-managed `.githooks/pre-push` hook, which runs `pwsh .\scripts\pre-push-quick-checks.ps1` automatically before each push — an incremental build plus the container-free fast test slice, not full coverage. CI runs the full coverage + format gate on every non-docs PR, so the hook is a quick local sanity check rather than the authoritative gate. Run `pwsh .\scripts\run-coverage.ps1` yourself before you mark a PR ready if you want the full local confidence check.
```
s205109: **Medium: this still makes a mandatory gate step optional.** The canonical five-step gate requires Coverage before ready, but `if you want the full local confidence check` tells this page's reader it may be skipped. Remove the optional wording and state the same requirement as `testing-workflow.md:13-29`.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,7 +4,7 @@ Brief description of changes.
 
 ## Checklist
 
-- [ ] [Canonical pre-PR gate](../docs/development/testing-workflow.md#canonical-pre-pr-gate) passes (build, format, coverage, `git diff --check`)
+- [ ] [The canonical gate](../docs/development/testing-workflow.md#canonical-pre-pr-gate) passes (build, format, coverage, `git diff --check`) — tick this before you mark the PR ready, not when you open the draft
```
s205109: **Medium: the checklist's gate summary omits the PowerShell-suite step.** The linked canonical gate has five steps and includes `pwsh .\scripts\test-fast.ps1 -Mode PowerShell`, but this parenthetical lists only four. A contributor can follow and tick the visible checklist while skipping a required suite. List all five steps or remove the incomplete summary.

docs/development/workflow.md
```diff
@@ -0,0 +1,771 @@
...
+- The same process drawn as a decision tree: [`workflow.html`](workflow.html)
+- One printable page: [`ahkflow-workflow-cheatsheet.html`](ahkflow-workflow-cheatsheet.html)
+  and [`ahk-workflow.pdf`](ahk-workflow.pdf)
+
+---
+
+## 1. The stage spine
+
+Stage names are fixed. The exit strings below are canonical. They appear word for word in
+`workflow.html`, in the cheatsheet, and in this table. A check compares all three.
```
s205109: **Medium: this claims a durable parity check already exists.** The only current comparison was a one-off script in the private implementation plan; the repository check remains unchecked wave-2 work in backlog 072. A later editor can trust this sentence and leave the HTML/PDF sources stale. Say that parity is manual until wave 2, or land the check now.

.claude/CLAUDE.md
```diff
@@ -13,51 +13,27 @@ concise, sacrifice grammar for brevity.
...
+### Plan before you edit
 
-Picking up a `backlog/` item is never trivial. Neither is any change to user-facing wording,
-labels, or names — a wrong shared assumption about a word is cheap to catch up front and expensive
-to undo later. Staying inside one file does not make such a change trivial. When a change matches
-both the exemption list and this paragraph, this paragraph wins: plan first.
+- Classify the change by Difficulty before you touch code. `complex` goes to Design, `moderate` to Plan, `trivial` straight to Execute — see [workflow.md#stage-1-pickup](../docs/development/workflow.md#stage-1-pickup).
+- A change is `trivial` only when all three predicates are provably false: more than one file changes, an interface other code depends on changes, app-facing text changes — see [workflow.md#stage-1-pickup](../docs/development/workflow.md#stage-1-pickup).
+- Repository documentation is not app-facing text, so a docs-only change can stay `trivial` — see [workflow.md#stage-1-pickup](../docs/development/workflow.md#stage-1-pickup).
+- Picking up a `backlog/` item is never `trivial` — see [workflow.md#stage-0-intake](../docs/development/workflow.md#stage-0-intake).
```
s205109: **Medium: this rule contradicts the canonical trivial path.** `workflow.md:505-515` allows a backlog item to carry `Difficulty: trivial` and route straight to Execute, while this index says that can never happen. Since `workflow.md` wins disagreements, this retained rule sends Claude down a different route. Remove it or make the canon explicitly prohibit trivial tracked items.

backlog/071-development-process-artifacts.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Acceptance criteria
+
+- [ ] `docs/development/workflow.md`: 11 stage nodes with all 8 fields, transition table
+      (success, failure, blocked, not-applicable, resume edges), current-stage field rule,
+      four worked walkthroughs
+- [ ] `docs/development/workflow.html`: self-contained decision tree, prints cleanly,
+      keyboard reachable, colour never the only signal; published as an Artifact with the
+      URL recorded
+- [ ] `docs/development/ahkflow-workflow-cheatsheet.html` and `docs/development/ahk-workflow.pdf`,
```
s205109, line 32: **Medium: the PDF freshness mechanism is not reproducible across checkouts.** The sidecar matches the LF source (`CC0FB5...1682`), but the same text as CRLF hashes to `4253CF...7ECE`. This repo has `core.autocrlf=true`, and `git check-attr` reports no HTML `eol` rule. Backlog 072 records the known defect, but this wave's own reproducibility criterion remains unmet. Normalize before hashing and add the HTML line-ending rule now, or leave this criterion open.

docs/development/workflow.md
```diff
@@ -0,0 +1,771 @@
...
+| resume | `gh pr view` for the live state, continue from ready-flip or merge | stay |
+
+<a id="stage-10-cleanup"></a>
+
+### Stage 10 — Cleanup
+
+- **Entry** — pull request merged
+- **Who** — Sonnet, default effort
+- **Technique** — worktree removal scripts, then session end
+- **Action** — remove the worktree and branch, clear context, update memory. A local holder process, such as a second terminal in the folder, is a failure to retry, not a blocked edge
```
s205109, line 532: **Medium: Cleanup has no executable success handoff.** The removal hook starts a detached watcher and returns; the watcher can remove the worktree only after the current `claude.exe` exits (`remove-worktree-local-dev.ps1:9-28`). That session therefore cannot run the success checks at lines 425-442, while the Stage field freezes at `9-ship` and no later session is named. Define a durable pending-cleanup record and a later confirmer, or provide a synchronous external cleanup path.

docs/development/workflow.md
s205109: Confirmed and fixed. Ship now pushes its closure commit before the ready flip: `workflow.md:382` (Action), `:419` (the rule and why it matters), `:431` (success edge). Push boundaries widened from stages 1-8 to 1-9 at `:539`. Mirrored in `workflow.html`. Commit `bbffcba`.

This overturns spec §7's "Ship success never pushes", so the spec needs the matching edit — flagged to the human, since the guard blocks me from committing to the private plans repo.

docs/development/workflow.md
s205109: Confirmed, and reproduced before fixing rather than taken on trust: `gh pr view --json body -q .body` returned `System.Object[]` with 9 elements and `[regex]::Matches` found 0 matches. Joined into one scalar it finds 1 and replaces correctly.

Fixed at `workflow.md:113` with the join plus a comment saying why it is not cosmetic. A note at `:129` records that the trailing `Select-String` needs no join, because it takes pipeline input line by line. Commit `bbffcba`.

docs/development/workflow.md
s205109: Confirmed. `new-worktree.ps1:221` throws without `-Name`, and both the branch convention and the discovery rule need the number. My round-1 note created this circularity.

Rather than invent a reservation scheme I documented the route that actually works today — file and commit the item on `main` first, then create the worktree — at `workflow.md:201`, and said plainly that backlog 071 itself did not follow it and should not be copied, because its number was picked by hand. Filed backlog 080 for the automation, including the concurrency requirement. Commits `bbffcba`, `1fe1c2e`.

docs/development/workflow.md
s205109: Confirmed — I hit this exact refusal twice while doing this work. Added `workflow.md:530`: the plans-repo commit runs from the main checkout, this applies to every agent and not only Claude, and backlog 076 removes the detour. Until then the rule is to state the handoff rather than retry the refused command. Commit `bbffcba`.

backlog/031-cli-winget-distribution.md
s205109: Confirmed. Checked: no 031 worktree, no local or remote branch, no PR, no progress file — so `4-execute` named a stage whose resume edge had nothing to read. Restamped `1-pickup` and given `Difficulty: moderate` (`031-cli-winget-distribution.md:10-11`). Commit `1fe1c2e`.

This changes a stamp value spec §7 named explicitly, so the spec needs the matching edit.

backlog/042-key-rebinds-as-first-class-rows.md
s205109: Confirmed — all three migrated items had zero `Difficulty` lines, which is why the fresh-session subject could not resolve Pickup's target. Added: 042 `complex`, 070 `moderate`, 031 `moderate` (`042:9`, `070:5`, `031:10`). Commit `1fe1c2e`.

docs/development/workflow.md
s205109: Confirmed — all six `blocked/` mentions moved items in, none brought them back. Added the move out at `workflow.md:507`: the `git mv` command, the removal of the unblock note, the Stage field left untouched, and the ordering rule that this happens *before* the resume edge. It also states that the folder is the durable record, so an item left in `backlog/blocked/` still reads as blocked whatever its Stage says. Commit `bbffcba`.

docs/development/workflow.md
s205109: Confirmed. The resume row now requires the whole exit condition rather than Difficulty alone, and names the placeholder-summary case that keeps the item in Intake — `workflow.md:186`, mirrored in `workflow.html`. Commit `bbffcba`.

docs/development/workflow.md
s205109: Confirmed. `cross-agent-git-guardrails.md:160` refuses sibling-worktree writes outright, so there is no direct file move between worktrees. The uncommitted route now says to commit in the round first and then cherry-pick, so both cases use the same git-reachable path — `workflow.md:618`. Commit `bbffcba`.

docs/development/workflow.md
s205109: Confirmed. Pickup now pushes and opens the pull request first, and stamps the Difficulty-jump Stage only after that succeeds. Until the stamp lands the item keeps `Stage: 1-pickup` and the worktree is the pointer, so a failed `gh pr create` leaves no split state — `workflow.md:228`, mirrored in `workflow.html`. Commit `bbffcba`.

docs/development/coverage.md
s205109: Confirmed. The optional wording is gone. The line now says coverage is step 4 of the canonical gate and is required, not an optional extra — `coverage.md:41`. Commit `24bac35`.

.github/PULL_REQUEST_TEMPLATE.md
s205109: Confirmed. The checklist now names all five steps, including the PowerShell suites that were missing — `.github/PULL_REQUEST_TEMPLATE.md:7`. Commit `24bac35`.

docs/development/workflow.md
s205109: Confirmed — that sentence claimed a check this repository does not have. It now says keeping the three representations in step is manual until wave 2, and tells the editor to change all three files in the same commit and regenerate the PDF — `workflow.md:27`. Commit `bbffcba`.

.claude/CLAUDE.md
s205109: Confirmed. Resolved in favour of the unit contract rather than by deleting the rule: `workflow.md:586` now states that a filed backlog item is never `trivial`, because `trivial` classifies work that runs as a housekeeping round and a round files no item. The `trivial` row in the Difficulty table therefore describes unfiled work only. `.claude/CLAUDE.md:24` now carries the same reason instead of a bare assertion. Commit `bbffcba`.

This settles a question spec §6 left open, so the spec needs the matching edit.

backlog/071-development-process-artifacts.md
s205109: Confirmed, and fixed here rather than left open. Added `*.html text eol=lf` to `.gitattributes`. `git check-attr` now reports `eol: lf` for the cheatsheet, and the live hash matches the committed sidecar `CC0FB549...1682` — the same value you quoted. The wave-2 check keeps LF-normalized hashing regardless, so it stays correct rather than merely consistent on one machine. Commit `24bac35`.

This widens the wave-1 allowlist a third time; recorded in backlog 071's notes with the reason.

docs/development/workflow.md
s205109: Confirmed. `remove-worktree-local-dev.ps1:9-28` spawns a detached watcher that outlives `claude.exe`, so the triggering session cannot run the three success checks — the folder is still locked while it is alive.

Cleanup now has two named endings, and the session says which it took: synchronous, run from a session outside the worktree, which checks all three conditions itself; or deferred to the watcher, where the next session in the repository runs the checks first. The merged pull request plus the item in `backlog/done/` are the pending-cleanup record, since no Stage field is written after the merge. Neither route lets a session claim success it did not observe. `workflow.md:450`, mirrored in `workflow.html`. Commit `bbffcba`.

s205109:
## Review round 2 — all 16 threads addressed

Every finding was verified against the code before anything was edited. **All 16 confirmed, nothing pushed back.** The four you flagged as blocking were all real, and the Critical one would have lost data.

| # | Finding | Fix |
|---|---|---|
| Critical | Ship closure commit never pushed | `workflow.md:382, :419, :431, :539` — Ship pushes before the ready flip; push boundaries now stages 1-9 |
| High | Multiline PR body breaks the Stage writer | `workflow.md:113` — joined to one scalar; reproduced `System.Object[]`/0 matches first |
| High | Intake needs the number before the script assigns it | `workflow.md:201` — documents the main-checkout route; backlog 080 filed for the automation |
| High | Plans-repo commit blocked from the worktree | `workflow.md:530` — names the main-checkout handoff until backlog 076 lands |
| Medium | 031 stamped at an unresumable stage | `031:10-11` — `1-pickup`, `Difficulty: moderate` |
| Medium | Migrated items lack Difficulty | `031:10`, `042:9`, `070:5` |
| Medium | No move out of `backlog/blocked/` | `workflow.md:507` — the move, and it runs before the resume edge |
| Medium | Intake resume bypasses its exit condition | `workflow.md:186` — the whole exit condition, not Difficulty alone |
| Medium | Sibling-worktree move is refused | `workflow.md:618` — commit first, then cherry-pick |
| Medium | Stage published before the PR exists | `workflow.md:228` — stamp only after `gh pr create` succeeds |
| Medium | Coverage made optional | `coverage.md:41` |
| Medium | Checklist listed 4 of 5 gate steps | `.github/PULL_REQUEST_TEMPLATE.md:7` |
| Medium | Parity check overclaimed | `workflow.md:27` — manual until wave 2 |
| Medium | Trivial rule contradicts the canon | `workflow.md:586` — a filed item is never `trivial` |
| Medium | PDF hash not portable | `.gitattributes` — `*.html text eol=lf`; `check-attr` now reports `eol: lf` |
| Medium | Cleanup cannot confirm itself | `workflow.md:450` — two named endings, neither claiming unobserved success |

Commits: `bbffcba` (process fixes), `24bac35` (gate completeness and hash portability), `1fe1c2e` (migrated items, backlog 080).

### Answering the review body

You noted the representations are structurally consistent but the process defects are real. That matches what I found: parity was green throughout and never caught any of this, because parity compares the three files to each other, not the process to the tooling. Every one of the four blocking findings came from checking a document claim against a script — `new-worktree.ps1:221`, the guard, the detached watcher, and how GitHub merges. That is the gap wave 2's drift guard should close, and it is worth saying out loud rather than filing quietly.

### Three things that need your decision

Four fixes overturn settled spec decisions, and the guard blocks me from committing to the private plans repo:

1. **Spec §7 "Ship success never pushes"** is now wrong — Ship must push its closure commit.
2. **Spec §7's stamp values** for 031 (`4-execute`) are now `1-pickup`.
3. **Spec §6** left open whether a filed item can be `trivial`; the canon now says it cannot.

The spec needs those three edits to stop contradicting `workflow.md`.

### Scope note

This round widened the wave-1 file allowlist a third time, to `.gitattributes`, so that this wave's own reproducibility criterion is actually true rather than deferred. Recorded in `backlog/071`'s notes with the reason, alongside the earlier `testing-workflow.md`, `coverage.md`, and PR-template widenings.

### Gate

All five steps green after the fixes: build 17 projects / 0 warnings, format clean, 18 PowerShell suites passed, coverage line 94.6% and branch 82.6% with every per-assembly threshold met, `git diff --check main...HEAD` and the bare form both clean. Parity all green, sidecar fresh, PDF one page, 8 stage anchors used with 0 broken.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+Ship flips the pull request to ready and closes the records before CI runs. So both
+non-success edges must undo what Ship already did.
+
+Failure converts the pull request back to draft
+(`gh pr ready <number> --undo --repo s205109/AHKFlowApp`). It returns to a pre-Review
+stage, and Review's entry condition is a draft pull request. Without the undo, that path is
+shut. Blocked keeps the ready pull request, because it resumes at Ship, and Ship needs a
+ready pull request.
+
+**Every Ship transition that makes a commit pushes it, success included.** GitHub merges
```
s205109, line 495: **Critical: the executable private plan still contains the Ship data-loss bug.** `docs/superpowers/plans/2026-08-10-development-process-plan-071.md:895-905` still says Ship success never pushes and that the merge carries the closure commit. The plan is the current task's durable execution record, and `.claude/CLAUDE.md` requires review-changed plans to be updated. Updating only the spec leaves a fresh Ship session able to delete the sole closure records again. Update and commit both private artifacts before Ship.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+
+The round pull request is created at Execute close with a `Stage: 5-simplify` line in its
+body. Exactly one such line exists from then on.
+
+```powershell
+$pr   = <round PR number>
+$next = '<N-name>'                      # e.g. 6-verify
+$rx   = '(?m)^Stage: [^\r\n]+'          # CRLF-safe: does not eat the carriage return
+
+# -join is not cosmetic. PowerShell captures multiline native output as System.Object[],
```
s205109, line 123: **High: the private plan still carries the reproduced `System.Object[]` failure.** Its normative round writer at `plans/2026-08-10-development-process-plan-071.md:745-759` still assigns native `gh` output directly to `$body`, so `[regex]::Matches` sees the array and finds zero Stage lines. Mirror this scalar join into the committed private plan, not only the public transcription.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+| resume | re-run `/simplify` on the diff; it is repeatable | stay |
+
+<a id="stage-6-verify"></a>
+
+### Stage 6 — Verify
+
+- **Entry** — code settled
+- **Who** — Sonnet, default effort
+- **Technique** — the AGENTS.md Verification routing table (`dck-verify`)
+- **Action** — produce the verification artifact the table names, then run the canonical five-step gate in [`testing-workflow.md`](testing-workflow.md#canonical-pre-pr-gate). A wait on the human for manual steps keeps the stage in progress; it is not an edge. An AGENTS.md exemption replaces the **artifact** only — for exemption 1, targeted text checks plus diff review stand in for a test. The gate still runs. It runs on a docs-only branch too, because CI skips its .NET steps there, so the local gate is the only .NET check that branch gets. Name the exemption, state the verdict, and take the success edge
```
s205109, line 417: **High: the private plan still permits exempt changes to skip the five-step gate.** `plans/2026-08-10-development-process-plan-071.md:849-855` says exemption work is both the artifact and the gate. That directly contradicts this corrected action and can let a resumed session run only text checks. Update the private plan so exemptions replace only the artifact.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+| not applicable | entered, nothing moved — 'nothing to document'; the verdict is mandatory, silence is not | 8-review |
+| resume | diff docs against the change, continue | stay |
+
+<a id="stage-8-review"></a>
+
+### Stage 8 — Review
+
+- **Entry** — work presentable, pushed at the last stage boundary, pull request still draft
+- **Who** — `--model opus --effort high`; Codex locally posts to the same pull request threads
+- **Technique** — two halves. **Producing** the review: a reviewer session, or Codex, reads the branch and posts threads. `custom-review-findings` cannot do this half — it only consumes findings. **Handling** the review: `custom-review-findings` (PR mode)
```
s205109, line 458: **High: the private plan can still complete Review without receiving a review.** `plans/2026-08-10-development-process-plan-071.md:875-881` names only `custom-review-findings`, omits the producer half, and exits on resolved threads without `Review received`. An empty thread list can therefore pass. Mirror this two-half technique and exact exit into the private plan.

PLAN-PROGRESS.md
```diff
@@ -0,0 +1,21 @@
...
+Document | - | verdict: nothing to document - the change set IS the documentation. Checked, not assumed: only AGENTS.md, testing-workflow.md and workflow.md describe the process, and all three were updated this round; no docs index exists that would need the six new files added; no document still states the old gate or review timing | known exception: the two dck-verify SKILL.md copies keep the pre-PR wording - skills are out of scope for wave 1 by 071 Out of scope | not-applicable edge -> 8-review
+Review round 2 | bbffcba 24bac35 1fe1c2e | 16 threads, all CONFIRMED, 0 pushback, 0 unclear, 0 deferred | 1 critical (Ship closure commit never pushed - would have lost the closure records at merge), 3 high, 12 medium | new: backlog 080 race-safe Intake | gate: all five steps green, parity all green, sidecar fresh, check-attr eol=lf | 16/16 replied and resolved, one summary comment posted | spec edits still owed: section 7 Ship push rule, section 7 stamp value for 031, section 6 trivial-item question
+Review failure edge: 071 8-review -> 4-execute (findings needed code changes)
```
s205109, line 21: **High: the current durable records disagree about where 071 stands.** This last progress line says the failure edge set 071 to `4-execute`, while `backlog/071-development-process-artifacts.md:9` says `8-review`; there are no later Execute, Simplify, Verify, or Document transition records. The Stage field wins, but `tracking current` is false and the failure transition was recorded after its fixes. Record the failure before fixing, then append the complete return path to Review.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+
+<a id="stage-0-intake"></a>
+
+### Stage 0 — Intake
+
+- **Entry** — friction, idea, or bug arrives
+- **Who** — Sonnet, default effort
+- **Technique** — `scripts/new-backlog-item.ps1`
+- **Action** — file the item where the work will run (see below), set Difficulty; for bugs, attach root-cause evidence or mark it `to-be-determined`
+- **Exit** — Item filed with the script, Difficulty set
```
s205109: **High: Intake still has two incompatible exit definitions.** The canonical Exit and success row require only filing plus Difficulty, but the failure and resume rows say a missing or placeholder summary must stay in Intake. A generated item with Difficulty and the untouched placeholder both passes and fails. Add summary completion to the canonical exit and all representations, or remove it from the failure/resume gates.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+- **Action** — produce the verification artifact the table names, then run the canonical five-step gate in [`testing-workflow.md`](testing-workflow.md#canonical-pre-pr-gate). A wait on the human for manual steps keeps the stage in progress; it is not an edge. An AGENTS.md exemption replaces the **artifact** only — for exemption 1, targeted text checks plus diff review stand in for a test. The gate still runs. It runs on a docs-only branch too, because CI skips its .NET steps there, so the local gate is the only .NET check that branch gets. Name the exemption, state the verdict, and take the success edge
+- **Exit** — Verification artifact green, gate green, verdict stated
+- **Next** — `7-document`
+- **Context** — safe to clear after green
+- **Review** — not yet reviewed
+
+| Edge | Condition | Target |
+|---|---|---|
+| success | exit condition met | 7-document |
+| failure | red; carry the failing output back | 4-execute |
```
s205109: **High: a Verify failure has no durable recovery task.** After all planned tasks have progress lines, this edge sets `Stage: 4-execute`, but Execute resume only looks for the first task without a progress line (`:311`). If the session ends after the transition, neither the failing command/output nor pending repair exists in `PLAN-PROGRESS.md`. Require the failure transition to append the red evidence and a recovery task before returning to Execute.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+- **Entry** — pull request merged
+- **Who** — Sonnet, default effort
+- **Technique** — worktree removal scripts, then session end
+- **Action** — remove the worktree and branch, clear context, update memory. A local holder process, such as a second terminal in the folder, is a failure to retry, not a blocked edge
+- **Exit** — Worktree gone, branch deleted, memory updated
+- **Next** — `terminal`; the `done/` location plus the merged pull request are the durable record, and the Stage field is never written after merge
+- **Context** — clear freely
+- **Review** — reviewed
+
+**The session that starts Cleanup usually cannot finish it.** The removal hook spawns a
```
s205109, line 560: **High: the normal Cleanup hook tests against stale local `main`.** `remove-worktree-local-dev.ps1:314-322,624-626` requires worktree HEAD to be an ancestor of local `main`. `gh pr merge` merges on GitHub and does not advance local `main`, so the hook normally preserves the just-merged worktree as “not merged”. A worktree session cannot update main under the guard. Assign a main-checkout synchronization step before removal, or make the script test a freshly fetched remote base.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+- **Next** — `3-plan`
+- **Context** — keep; it feeds the plan
+- **Review** — spec reviewed before Plan
+
+| Edge | Condition | Target |
+|---|---|---|
+| success | exit condition met | 3-plan |
+| failure | spec rejected; revise with the objections | stay |
+| blocked | external decision or platform gap | blocked/ |
+| not applicable | reclassification at entry: no spec needed; set Difficulty `moderate`, say so; a reclassified `trivial` continues through Plan's not-applicable edge | 3-plan |
```
s205109: **High: this edge still permits the tracked `trivial` path the new rule forbids.** Section 5 says every filed item remains tracked and gets a plan even when small (`:586-592`), but Design can still reclassify it to `trivial`, and Plan's not-applicable edge at `:289` then skips the committed plan. A `to-be-determined` item now has two incompatible legal paths. Remove the tracked-trivial branches and use `moderate` for a filed item.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+stranded.
+
+There is a second constraint, and the two together decide the route. The worktree needs the
+item's **number before it exists**: `scripts/new-worktree.ps1:221` throws without `-Name`,
+the branch convention is `feature/wt-NNN-<topic>`, and the pre-PR discovery rule below finds
+work by its number-prefixed worktree. But `scripts/new-backlog-item.ps1` only assigns the
+next free number when it writes the file. Filing inside a worktree that is already named
+therefore needs a number nobody has assigned yet.
+
+**The route today: file and commit the item on `main` first, then create the worktree.**
```
s205109, line 211: **Medium: the manual Intake route commits an item without its required Stage record.** The current template and `new-backlog-item.ps1` add neither Difficulty nor Stage. The code block commits the generated file, while Pickup assumes it already carries `Stage: 1-pickup`. Until wave 2 changes the script, explicitly edit the summary, Difficulty, and `- **Stage**: 1-pickup` before this commit.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+A housekeeping round files no item at all, so none of this applies to it.
+
+<a id="stage-1-pickup"></a>
+
+### Stage 1 — Pickup
+
+- **Entry** — item chosen
+- **Who** — Sonnet, default effort
+- **Technique** — `scripts/new-worktree.ps1` (`-BaseRef` for stacked work), or the housekeeping worktree
+- **Action** — choose the execution location by Difficulty: a dedicated worktree for `moderate` and `complex`, the housekeeping worktree for `trivial`. Confirm branch and base, and state both. For a tracked item: push the branch, open the draft pull request, and **only then** stamp the Stage the Difficulty jump names — publishing the next Stage before the pull request exists would claim an exit condition that is still false. Until that stamp lands the item keeps `Stage: 1-pickup`, and the worktree is the pointer. For a housekeeping round: no item, so no Stage stamp and no push here. The round worktree is the pointer, and the route is the round pull request opened at Execute close
```
s205109, line 257: **Medium: the corrected Pickup ordering requires a second push but never states it.** The first push exists only so `gh pr create` can run; the Stage transition commit happens afterward. Without an explicit commit and second push, the remote item remains `1-pickup` while local state advances. State both pushes and distinguish the branch-publication push from the stage-completion push.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+field in the same commit that completes the stage. A failure edge sets the field back to
+the target stage. A blocked item keeps its last stage; the move to `backlog/blocked/` plus
+the unblock note carry the rest.
+
+**Coming back out of `backlog/blocked/`.** Every blocked edge moves an item in; nothing
+moves it back, so the move out is a step of its own and it comes **before** the resume
+edge. When the external condition the unblock note names has cleared:
+
+```powershell
+git -C <worktree> mv backlog/blocked/<NNN>-<slug>.md backlog/<NNN>-<slug>.md
```
s205109: **Medium: this global unblock route assumes a worktree that does not always exist.** Intake can take its blocked edge before Pickup creates a worktree, and the new Intake route stores that item on `main`. `<worktree>` therefore cannot resolve for every blocked edge as claimed. Use the checkout that owns the item and state the main-session handoff for pre-Pickup blocks.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+# remove the unblock note, keep the Stage field exactly as it was
+git -C <worktree> commit -m "chore: unblock <NNN>, <what cleared>"
```
s205109: **Medium: removing the unblock note is not included in this commit.** `git mv` stages the rename and its current contents. Editing the destination afterward changes only the worktree, so plain `git commit` records the old unblock note and leaves the removal unstaged. Add `git add backlog/<NNN>-<slug>.md` after the edit.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+
+One round walks Intake once (not applicable, because a round has no item), then Pickup,
+then Execute, then the verdict stages, then Review, Ship, and Cleanup. Changes accumulate
+inside Execute with one commit each. The round closes when a few have gathered. Then it
+pushes for the first time and opens the round pull request.
+
+A single trivial change is Execute-internal. It takes no stage edges of its own.
+
+A change that grows in importance does not close the round. It is reclassified and leaves
+before the round pushes. Move its files into the new worktree, or cherry-pick its commit
```
s205109: **Medium: this summary still directs the sibling-worktree write the detailed fix forbids.** Lines 618-624 correctly require commit then cherry-pick because the guard rejects direct sibling writes, but this earlier instruction still says to move the files. The private spec repeats the same stale route at §6 lines 166-169. Replace both with the git-reachable route.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+### The housekeeping round is the stage-machine unit for trivial work
+
+One round walks Intake once (not applicable, because a round has no item), then Pickup,
+then Execute, then the verdict stages, then Review, Ship, and Cleanup. Changes accumulate
+inside Execute with one commit each. The round closes when a few have gathered. Then it
+pushes for the first time and opens the round pull request.
+
+A single trivial change is Execute-internal. It takes no stage edges of its own.
+
+A change that grows in importance does not close the round. It is reclassified and leaves
```
s205109, line 102: **Medium: the testing guide gives the opposite escalation action.** `testing-workflow.md:9` says to open the housekeeping PR when one change grows in importance. This canonical rule says growth does not close the round and the change must leave before the round pushes. Following the testing guide publishes non-trivial work in the housekeeping PR and skips its tracked workflow. Make the guide point here instead.

s205109 (commented):
Third-round cold review of the complete PR. Round 2 is not fully closed: 10 original findings are fixed and 6 remain partial. This pass found 23 actionable threads: 1 Critical, 8 High, and 14 Medium. The three already acknowledged private-spec edits are not counted. CI is green and structural parity is green, but the executable private plan still contains the Ship data-loss instruction. Verdict: not ready to merge.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+| Resume | Read the `Stage` field, re-run that stage's entry check | Read the round pull request body's `Stage:` line, re-run that stage's entry check. Before the pull request exists, read the round worktree's commit log |
+| Blocked | The item moves to `backlog/blocked/` with the unblock note | Two kinds. A single change that blocks is filed as its own item, leaves the round, and moves to `backlog/blocked/`; the round continues. A round-level blocker belongs to no single change: file an item naming the round pull request and the blocker, move that item to `backlog/blocked/`, and stop the round |
+| Records at Ship | Boxes ticked, item moved to `backlog/done/`, progress file deleted | Nothing extra |
+
+### The housekeeping round is the stage-machine unit for trivial work
+
+One round walks Intake once (not applicable, because a round has no item), then Pickup,
+then Execute, then the verdict stages, then Review, Ship, and Cleanup. Changes accumulate
+inside Execute with one commit each. The round closes when a few have gathered. Then it
+pushes for the first time and opens the round pull request.
```
s205109, line 89: **Medium: round close is not durable between push and PR creation.** If the first push succeeds and `gh pr create` fails, the only record is the same commit log used while changes are still accumulating. Execute resume says to continue committing, so a fresh session cannot tell that close was decided and PR creation is pending. Record close intent before the push, or require resume to inspect upstream and PR state and retry creation.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+
+A round has no item, no progress file, and nothing extra to close at Ship. Its non-success
+edges restore no records. They rewrite the `Stage:` line in the round pull request body
+instead, and they push only if the recovery work itself makes a commit.
+
+| Edge | Condition | Target |
+|---|---|---|
+| success | closure commit pushed, then CI green and merged | 10-cleanup |
+| failure | CI red after ready; convert the PR back to draft first. Tracked item: `git mv` the item back out of `backlog/done/`, restore `PLAN-PROGRESS.md`, set `Stage: 6-verify` — one commit — then push. A round: rewrite the PR body's `Stage:` line to `6-verify`; no records to restore, no transition commit, so no push. Start recovery only after that | 6-verify |
+| blocked | merge depends on something outside the repository, such as a required-check outage; the PR stays ready. Tracked item: restore `PLAN-PROGRESS.md`, `git mv` the item from `backlog/done/` to `backlog/blocked/` with the unblock note, keep `Stage: 9-ship`, commit, push. A round has no item: file one naming the round PR and the blocker, move that item to `backlog/blocked/`, and leave the round PR body at `Stage: 9-ship` | blocked/ |
```
s205109: **Medium: a round-level blocker item has no closure path after the round resumes.** Unblocking moves its marker from `backlog/blocked/` into open `backlog/`, but round Ship closes “nothing extra” at lines 425-427. When the outage clears and the round merges, that blocker-only item remains open forever. Define how successful round Ship moves that marker to `done/` or otherwise closes it.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+process exits (`scripts/remove-worktree-local-dev.ps1:9-28`). So the session that triggers
+removal cannot run the three success checks below — the folder is still locked while it is
+alive.
+
+Cleanup therefore ends in one of two ways, and the session says which:
+
+1. **Synchronous.** Removal runs from a session outside the worktree — the main checkout,
+   or any shell not standing in the folder. That session runs all three checks itself and
+   takes the success edge.
+2. **Deferred to the watcher.** The session triggers removal, states that the watcher owns
```
s205109: **Medium: these permanent records cannot serve as a pending-Cleanup marker.** The merged PR and `backlog/done/` item look identical before and after the three checks, and a housekeeping round has no done item at all. A later session has no trigger telling it which merged work needs confirmation “before anything else”. Persist a distinct pending record, or make a repository-start sweep discover and verify all unremoved merged worktrees.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+| success | location chosen, base stated, PR in place; jump by Difficulty | 2-design/3-plan/4-execute |
+| failure | wrong base discovered; recreate the worktree with `-BaseRef` | stay |
+| blocked | prerequisite outside the repository; for a round, the one blocked change is filed as its own item and only that item moves | blocked/ |
+| not applicable | cannot occur: every item chooses a location | none |
+| resume | worktree exists; re-confirm branch and base, then continue at the recorded Stage for a tracked item, or from the round's commit log and PR state for a round | stay |
+
+**The draft pull request opens before any gate has run, and that is deliberate.** The pull
+request exists from Pickup so it can point at the work through Design and Plan. It is a
+**draft**, which is not a request to merge. The canonical five-step gate in
+[`testing-workflow.md`](testing-workflow.md#canonical-pre-pr-gate) is therefore a
```
s205109, line 321: **Medium: the linked gate uses the wrong diff base for stacked PRs.** Stacked work is supported through `-BaseRef`, but `testing-workflow.md:28-31` hardcodes `main...HEAD`. That includes prerequisite-branch commits which are not part of the stacked PR and can fail it for unrelated whitespace. Derive the PR's actual base ref and run `<base>...HEAD`.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+
+### Safe to clear context
+
+Three rules, from two documented incidents:
+
+1. Keep context between [Design](#stage-2-design) and [Plan](#stage-3-plan). It holds
+   line-level knowledge the spec only summarizes.
+2. Clear it between [Plan](#stage-3-plan) and [Execute](#stage-4-execute). The plan carries
+   the line numbers.
+3. Never clear between Ship start and pull-request-open. That context writes the pull
```
s205109: **Medium: this context rule refers to an ordering that no longer exists.** The draft pull request opens at Pickup, long before Ship. “Never clear between Ship start and pull-request-open” is therefore either impossible or means retaining one context through the whole workflow. Name the actual Ship artifact that needs context, such as the final PR description, or remove this stale rule.

backlog/072-process-wave-2-parity-drift-guard-templates.md
```diff
@@ -0,0 +1,110 @@
...
+- [ ] The check builds the legal edge-target set from the 11 stage ids it extracts, plus
+      `2-design/3-plan/4-execute`, `stay`, `terminal`, `blocked/`, and `none`. Every edge
+      target must be a case-sensitive member of that set.
+- [ ] The check recomputes the SHA-256 of the cheatsheet HTML and compares it against
+      `docs/development/ahk-workflow.pdf.source.sha256`. A mismatch means the PDF is
+      stale. It also checks that the PDF page tree contains `/Count 1`.
+- [ ] **The hash is line-ending independent.** Both the check and the sidecar generator
+      read the file, replace every CRLF with LF, and hash the UTF-8 bytes of that
+      normalized text. Hashing the raw bytes makes the result depend on the checkout: this
+      repository sets `core.autocrlf=true` and `.gitattributes` carries no `*.html` rule,
```
s205109: **Medium: this future implementation contract contradicts its next checked criterion.** It says `.gitattributes` has no HTML rule, while lines 41-44 correctly record that round 2 added `*.html text eol=lf`. Wave 2 will read this as current state. Rewrite the example in past tense while keeping normalized hashing required.

docs/development/workflow.md
```diff
@@ -0,0 +1,849 @@
...
+- The same process drawn as a decision tree: [`workflow.html`](workflow.html)
+- One printable page: [`ahkflow-workflow-cheatsheet.html`](ahkflow-workflow-cheatsheet.html)
+  and [`ahk-workflow.pdf`](ahk-workflow.pdf)
+
+---
+
+## 1. The stage spine
+
+Stage names are fixed. The exit strings below are canonical. They appear word for word in
+`workflow.html`, in the cheatsheet, and in this table. **Keeping the three in step is
```
s205109, line 27: **Medium: the private spec still claims parity is machine-compared today.** Spec §3 lines 56-59 says the exit strings are canonical and machine-compared, while this corrected canon says comparison remains manual until wave 2. This is an additional spec edit beyond the three already listed in `PLAN-PROGRESS.md:20`.

backlog/071-development-process-artifacts.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Out of scope
+
+- Changes to `scripts/`, `.github/workflows/`, guards, hooks, `CONTEXT.md`, `docs/adr/`, or
+  any skill — later waves
+- Personal `~/.claude` configuration — final wave, kept out of the repository
+
+## Notes / dependencies
+
+- **The wave-1 file allowlist was widened by two files during review round 1**, with the
```
s205109, line 52: **Medium: the private plan still rejects the review-approved files documented here.** Its global allowlist at `plans/2026-08-10-development-process-plan-071.md:40-45` says “Nothing else”, excluding `testing-workflow.md`, `coverage.md`, the PR template, and `.gitattributes`. A resumed review-fix session can refuse required changes. Update the plan's allowlist with these recorded exceptions.

docs/development/workflow.md
s205109: Confirmed — verified at `plans/…-071.md:900`: "Ship's success is the only Ship transition that never pushes", and the success edge still reads "merged; no push — the merge carries it".

**Not fixed, because this session cannot write it.** `docs/superpowers` is a symlink into the main checkout and the guard refuses it from a worktree — the same gap backlog 076 tracks. Handed to the human with the exact edit, flagged as the highest-priority item of the round. The public canon at `workflow.md:419` is correct; the executable plan is not, and you are right that the plan is what a resumed Ship session actually follows.

docs/development/workflow.md
s205109: Confirmed at `plans/…-071.md:749` — `$body = gh pr view …` with no join, so `[regex]::Matches` sees `System.Object[]`. Same defect I reproduced and fixed in the public copy last round. **Not fixed here:** private repo, guard-blocked. Handed over with the exact replacement line.

docs/development/workflow.md
s205109: Confirmed at `plans/…-071.md:853` — "the exemption's own work **is** the artifact and the gate". **Not fixed here:** private repo, guard-blocked. Handed over. The public canon at `workflow.md:341` already says the exemption replaces only the artifact.

docs/development/workflow.md
s205109: Confirmed at `plans/…-071.md:878` — Technique names only `custom-review-findings`, and the Exit is "All review threads resolved" without `Review received`. **Not fixed here:** private repo, guard-blocked. Handed over with both the two-half technique and the corrected exit string.

PLAN-PROGRESS.md
s205109: Confirmed. The return path was missing entirely: `PLAN-PROGRESS.md` jumped from the round-2 failure edge to nothing. Appended the full Execute → Simplify → Verify → Document → Review path with its verdicts and gate results.

You are also right about the ordering, and it applied to both rounds: I recorded the failure edge *after* making the fixes. `workflow.md:350` now requires the failure transition to be written first, with the red evidence and a recovery task. I added an explicit NOTE in `PLAN-PROGRESS.md` saying rounds 2 and 3 did not follow that order, rather than quietly writing a tidy history. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — the canonical exit said "filed, Difficulty set" while the failure and resume rows gated on the summary, so a generated item with the untouched placeholder both passed and failed.

Took your first option: the canonical exit is now `Item filed with the script, summary written, Difficulty set`, updated in the spine table, the Intake node, the success row, `workflow.html`, the cheatsheet, and the regenerated PDF. Parity is green across all three. The spec's copy of the string is guard-blocked and is in the handover list. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — Execute resume looks for the first task with no progress line, so after every task has one, a Verify failure left nothing to find. The failure edge now requires appending the failing command, its output, and a named recovery task to `PLAN-PROGRESS.md` *before* the transition — `workflow.md:350`. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed, and this one is a genuine trap: `remove-worktree-local-dev.ps1:314-322` tests `merge-base --is-ancestor HEAD main` against **local** `main`, which `gh pr merge` never advances — so the normal post-merge path preserves the worktree as "not merged".

Added the main-checkout sync step (`checkout main` + `pull`) before removal at `workflow.md:450`, and noted that making the script test a freshly fetched remote base is backlog 073's business rather than a rule anyone can follow today. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — and this was introduced by my own round-2 fix, which added the never-trivial rule to section 5 without removing the two edges that contradict it.

Design's not-applicable row no longer offers `trivial` (`:268`), and Plan's not-applicable edge is now `none` with the reason that every filed item gets a committed plan however small (`:289`). The target change is mirrored in `workflow.html`, and parity is green. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — the script writes the template unchanged, so it supplies neither Difficulty nor Stage and leaves the placeholder summary. The block at `workflow.md:201` now names all three edits before the commit, and points at backlog 072 for removing two of them. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed. Added at `workflow.md:245`: the first push publishes the branch so `gh pr create` has something to target, the second carries the Stage transition commit, and without the second the remote item still reads `1-pickup` while local state has moved on. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — an item blocked at Intake has no worktree at all, and under the corrected Intake route it lives on `main`. The command now runs in the checkout that owns the item, with the main-session handoff named for pre-Pickup blocks — `workflow.md:512`. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed. `git add` added after the edit, with the reason stated: `git mv` stages the rename with the file's contents as they were, so an edit made afterwards stays unstaged and the commit would record the old unblock note — `workflow.md:514`. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — the section 2 summary still said "move its files into the new worktree" while section 5 correctly required commit-then-cherry-pick. The summary now states the git-reachable route and the reason (the guard refuses sibling writes) — `workflow.md:94`. The spec's copy at §6:166-169 is guard-blocked and is in the handover list. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed. `testing-workflow.md:9` now says a change that grows in importance leaves the round rather than closing it, and links here — publishing non-trivial work inside the housekeeping PR skips its whole tracked workflow. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — between the push and a successful `gh pr create` the commit log is indistinguishable from a round still accumulating, so resume would keep committing. Round close is now recorded before the push as an explicit marker commit, and round resume checks upstream and PR state before reading the log, retrying `gh pr create` when the marker exists and no PR does — `workflow.md:89`. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — "nothing extra" swallowed the one record a round can legitimately own. Ship now has an explicit exception: a round-level blocker item is `git mv`-ed to `backlog/done/` and ticked in the same closure commit Ship pushes — `workflow.md:433`. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — the merged PR and the `done/` item look identical before and after the checks, and a round has no `done/` item at all, so neither can trigger anything.

The trigger is now the leftover itself: a session starting in this repository lists worktrees, and for any whose branch is already merged it runs the three checks before its own work. A worktree that should be gone is the record that it is not gone yet — `workflow.md:461`. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed. `testing-workflow.md` now derives the real base with `gh pr view --json baseRefName` and runs `<base>...HEAD`, with the stacked-PR reason stated. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — another one introduced by my round-2 change: moving PR creation to Pickup made "never clear between Ship start and pull-request-open" describe an ordering that no longer exists. The rule now names the actual Ship artifacts that need context — the final PR description, the recap, and the closure records — at `workflow.md:690`. Commit `df6e018`.

backlog/072-process-wave-2-parity-drift-guard-templates.md
s205109: Confirmed — the example described the pre-round-2 state in the present tense, directly above the criterion recording that round 2 fixed it. Rewritten in past tense, with normalized hashing still required so the check stays correct rather than merely consistent on one machine — `backlog/072:36`. Commit `df6e018`.

docs/development/workflow.md
s205109: Confirmed — spec §3 still says the exit strings are machine-compared, while the corrected canon says comparison is manual until wave 2. **Not fixed here:** private repo, guard-blocked. Added to the handover list as a fourth spec edit beyond the three already recorded.

backlog/071-development-process-artifacts.md
s205109: Confirmed at `plans/…-071.md:40-45` — the allowlist ends "Nothing else", which now excludes four files this review approved. A resumed session would refuse required changes. **Not fixed here:** private repo, guard-blocked. Handed over with the four exceptions and the reason each was approved.

s205109:
## Review round 3 — 16 fixed here, 7 need your hands

All 23 verified against the source before any edit. **23 confirmed, 0 pushed back.** Commit `df6e018`; gate green.

Two of the 23 were **introduced by round 2's own fixes** — the tracked-`trivial` edges (I added the never-trivial rule to section 5 without removing the two edges contradicting it) and the stale context rule (moving PR creation to Pickup made it describe an ordering that no longer exists). Worth naming plainly rather than filing quietly.

### Fixed in this repository (16)

`workflow.md` unless stated: Intake exit gained `summary written` and now agrees with its own failure/resume gates (`:31, :175, :182`, plus `workflow.html`, cheatsheet, PDF) · Verify failure records red evidence and a recovery task before transitioning (`:350`) · Cleanup syncs local `main` first (`:450`) · Cleanup uses leftover worktrees as its pending record (`:461`) · tracked-`trivial` edges removed (`:268, :289`) · Intake manual route fills summary, Difficulty and Stage (`:201`) · Pickup states both pushes (`:245`) · unblock names the owning checkout (`:512`) and stages the edited file (`:514`) · sibling-move summary corrected (`:94`) · round close durable before push (`:89`) · round-level blocker closes at Ship (`:433`) · context rule names the real Ship artifacts (`:690`) · `testing-workflow.md` stops contradicting round escalation and derives the real base ref for stacked PRs · `backlog/072:36` rewritten in past tense · `PLAN-PROGRESS.md` return path restored.

### Needs the private plans repo (7) — including the Critical

`docs/superpowers` is a symlink into the main checkout and the guard refuses it from a worktree, so I cannot make these. That is backlog 076's gap, and it is now blocking the highest-severity finding of the round.

**In `plans/2026-08-10-development-process-plan-071.md`:**

| Line | Change |
|---|---|
| 900 | Delete "Ship's success is the only Ship transition that never pushes"; Ship pushes its closure commit before the ready flip |
| ~905 | Success edge: `merged; no push — the merge carries it` → `closure commit pushed, then CI green and merged` |
| 749 | `$body = gh pr view …` → `$body = (gh pr view …) -join "\n"` |
| 853 | Exemption "**is** the artifact and the gate" → replaces the **artifact** only; the gate still runs |
| 878 | Review technique: add the producer half; exit → `Review received, all threads resolved, gate re-green` |
| 40-45 | Allowlist "Nothing else" → add `testing-workflow.md`, `coverage.md`, `.github/PULL_REQUEST_TEMPLATE.md`, `.gitattributes` |
| §6 166-169 | Sibling-worktree move → commit then cherry-pick |

**In `specs/2026-08-10-development-process-design-071.md`:** §7 Ship push rule · §7 stamp value for 031 · §6 trivial-item question · §3 "machine-compared" → manual until wave 2 · Intake exit string gains `summary written`.

### The pattern is worth a decision

Findings per round: **9 → 16 → 23.** Each round has also introduced defects the next round found. The document is ~700 normative lines being edited in place under review pressure, and the two halves of the canon — public `workflow.md` and the private plan — drift apart every time only one is reachable.

I do not think a fourth round of the same shape converges. Two options worth considering:

1. **Land the guard exception (backlog 076) first**, then fix both halves in one pass. Most of round 3's severity came from the private half being unreachable, not from the rules being wrong.
2. **Split wave 1**: merge the artifacts that are stable (design system, cheatsheet, PDF, the AGENTS.md reduction) and keep `workflow.md` open on its own branch until the stage machine settles.

### Gate

Build 17 projects / 0 warnings · format clean · 18 PowerShell suites · coverage 94.6% line, 82.6% branch, all thresholds met · `git diff --check` clean both forms · parity all green · sidecar fresh · PDF one page.

PR #289, opened 2026-08-11
feat: plans-repo write exception for worktree sessions (backlog 076)

s205109:
Implements backlog 076. Spec: docs/superpowers/specs/2026-08-11-plans-repo-guard-exception-design.md


s205109:

s205109:
## Review findings addressed

Reviewer guide split into five checkable claims. One is a real defect and is fixed. Four do not
hold against the code at `1644fa35`. Every verdict below was checked by running the parser, not by
reading the snippet in the review.

| # | Finding | Verdict | Outcome |
|---|---|---|---|
| 1 | `mv --target-directory=<MAIN>/x file` from a worktree does not report the destination, so the guard is bypassed | Pushback | The attached long form is read at `scripts/agents/agent-worktree-guard.common.ps1:1492`. Probe: `mv --target-directory=../../x a.txt` returns `[../../x, a.txt]`. |
| 2 | `mv -t<MAIN>/x file` does not report the destination | Pushback | The attached short form is read by the cluster branch. Probe: `mv -t../../x a.txt` returns `[../../x, a.txt]`. |
| 3 | The test at line 1727 fails: the code returns `@('a.txt')` instead of `@('out', 'a.txt')` | Pushback | The suite was green at `1644fa35`. Probe: `mv --target-directory=out a.txt` returns `[out, a.txt]`. |
| 4 | `Get-AgentTargetDirectoryOption`'s cluster regex cannot extract `out` from `--target-directory=out` | Pushback | True but harmless: the `--target-directory=*` branch returns first, so the cluster regex never sees that token. |
| 5 | `mv -tout a.txt` matches `^-[a-zA-Z]*t$` and wrongly consumes `a.txt` as the `-t` value | Partial | Real defect, fixed. The diagnosis is off on two points, below. |

### The fix

`-tout` ends in `t`, so the bare-cluster regex `^-[a-zA-Z]*t$` matched it and the reader ate the
next token as the target directory. getopt does the opposite: a short cluster ends at its **first**
`t`, and everything after that letter is the option value. So `-tout` is `-t` with the value `out`.

`Read-AgentTargetDirectoryToken` (`scripts/agents/agent-worktree-guard.common.ps1:1489`) now decides
the spelling once, using a lazy quantifier so the match stops at the first `t`
(`agent-worktree-guard.common.ps1:1503`). Both `Get-AgentTargetDirectoryOption`
(`:1530`) and `Get-AgentMoveArgumentSet` (`:1568`) call it, so the two readers can no longer disagree.

Regression tests, red before the fix and green after:
`tests/AgentWorktreeGuard.Tests.ps1:1731`, `:1732`, `:1733`, `:1766`.

### Two corrections to finding 5

- **Not "zero targets reported".** The old code reported `a.txt` — it landed in the target-directory
  slot rather than the source slot. The destination `out` was the path that went missing.
- **No guard bypass was reachable.** To slip past the guard the value must name a path outside the
  worktree, which needs `/`, `\`, `:` or `.`. The faulty regex accepts letters only, so no such
  token could ever reach it. The defect is a correctness bug in the parser, not a security hole.

### Gate

- `pwsh .\scripts\test-fast.ps1 -Mode PowerShell` — 18 suites passed
- `-Mode Coverage` skipped: the branch changes no .NET file (`git diff --name-only main...HEAD`
  lists only Markdown, `scripts/agents/*.ps1`, and `tests/*.ps1`), so coverage cannot move.

No review threads existed on this PR, so there was nothing to resolve. The findings arrived as a
single bot comment, which is why they are all answered here.

PR #294, opened 2026-08-13
fix: stop concurrent test runs from destroying coverage data (backlog 082)

s205109:
## What

Backlog 082: two overlapping local test runs destroyed each other's coverage data, and the
resulting failure blamed coverage.

Three layers, in order of how much they prevent.

**1. A shared exclusive lock.** `scripts/test-run-lock.common.ps1` holds one `.test-run.lock`
per repository root with `FileShare::None`. `test-fast.ps1` and `run-coverage.ps1` both take
it. The second run to start fails immediately and names the first run's mode and process id.
The lock covers those two scripts only — a bare `dotnet test`, a bare `dotnet build`, and an
IDE build take no lock and can still collide.

**2. A completeness check.** `run-coverage.ps1` now runs each test project into
`TestResults/coverage/<ProjectName>/` and compares the coverage files produced against the
projects the solution names. When any project produced no `coverage.cobertura.xml`, the run
throws before ReportGenerator, so no number from an incomplete run is ever printed. This is
the safety net for every collision the lock cannot see.

**3. An honest gate message.** `check-coverage-thresholds.py` reports a missing assembly under
`::error title=Coverage input incomplete`, separate from the threshold error. A real
regression still prints `failed per-assembly coverage thresholds`.

## Root cause, measured

The **collector** holds the write, not the build and not the test execution. coverlet
instruments every module in a test project's output folder at test-session start: it copies
each module aside and writes an instrumented module back over it. `coverlet.runsettings` has
no `<Include>` filter, so `AHKFlowApp.Application.dll` inside `AHKFlowApp.UI.Blazor.Tests\bin`
is instrumented too — which is why the reported failure named a file that does not look like
it belongs to that project.

The dangerous part: `dotnet test` **exits 0** while coverlet writes **no** coverage file for
that project. The merged report then silently loses every assembly only that project covered,
and the gate calls it a coverage regression.

The item's second open question is also answered: no test ever executes against an assembly
the run did not produce. The run either loads the correct assembly and passes, or fails to
load it and reports a test failure. The withdrawn false-green claim stays withdrawn. The whole
defect is coverage-data loss.

## Verification

- `tests/CoverageRunLock.Tests.ps1` — 5 cases, pass.
- `tests/CoverageInputCompleteness.Tests.ps1` — 5 cases, pass.
- Live collision check: second run refused, named holder mode `Coverage` and process id.
- `pwsh .\scripts\test-fast.ps1 -Mode PowerShell` — all 20 suites pass.
- `dotnet build AHKFlowApp.slnx --configuration Release` — 17 projects, 0 errors, 0 warnings.
- `dotnet format AHKFlowApp.slnx --verify-no-changes` — clean.
- `pwsh .\scripts\test-fast.ps1 -Mode Coverage` — gate green, line 94.6%, branch 82.6%,
  identical to before the change.
- `git diff --check main...HEAD` — clean.

Per-project `dotnet test` cost: 431 seconds wall clock for the whole coverage path, gate
included. Acceptable, so the fallback to one solution-wide run is not needed.

## Notes

- `ci.yml` now pins the SDK from `global.json` and Python 3.12 in the `powershell-suites` job.
  The new suite is the first PowerShell suite that starts `dotnet` and calls `python`.
- CI runs its own `dotnet test` in `ci.yml`, not `run-coverage.ps1`, so CI gets the gate
  message change but not the completeness check. CI runners are single-run. Left alone on
  purpose.
- One acceptance criterion was corrected in the backlog item. Criterion 5 said the test must
  remove one expected TRX, which contradicted criterion 2: Coverage produces no TRX files. It
  now removes one expected per-project `coverage.cobertura.xml`.

Closes backlog 082 — item moved to `backlog/done/`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,6 +103,13 @@ jobs:
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v4
+      # tests/CoverageInputCompleteness.Tests.ps1 runs `dotnet sln list` and the coverage gate.
+      - uses: actions/setup-dotnet@v4
+        with:
+          global-json-file: global.json
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
       # The script runs every tests/*.Tests.ps1 as its own process, so one failing suite fails
       # this job and the rest still run. Listing the suites here let fifteen of them fail
       # silently - see backlog/done/066-ci-swallows-powershell-suite-failures-in-worktree-powershell-tests.md.
--- a/.gitignore
+++ b/.gitignore
@@ -461,6 +461,10 @@ coverage-report/
 CoverageReport/
 **/TestResults/
 
+# Shared test-run lock (scripts/test-run-lock.common.ps1)
+.test-run.lock
+.test-run.lock.owner
+
 # Superpowers brainstorming scratch (local-only)
 .superpowers/
 
--- a/backlog/082-concurrent-coverage-runs-corrupt-each-others-results.md
+++ b/backlog/082-concurrent-coverage-runs-corrupt-each-others-results.md
@@ -6,7 +6,7 @@
 - **Type**: Bug
 - **Interfaces**: none (test scripts)
 - **Difficulty**: moderate
-- **Stage**: 1-pickup
+- **Stage**: 3-plan
 
 ## Summary
 
--- a/backlog/082-concurrent-coverage-runs-corrupt-each-others-results.md
+++ b/backlog/082-concurrent-coverage-runs-corrupt-each-others-results.md
@@ -77,29 +77,42 @@ destroys the first's evidence mid-flight.
 
 ## Acceptance criteria
 
-- [ ] **A second concurrent run cannot start, or cannot collide — across scripts, not just
+- [x] **A second concurrent run cannot start, or cannot collide — across scripts, not just
       within one.** The lock must be shared by `run-coverage.ps1` and `test-fast.ps1`, since
       the observed collision was between them. A lock held only by `run-coverage.ps1` would
       not have prevented it.
-- [ ] The coverage gate refuses to report when its input is **incomplete**, measured in
+- [x] The coverage gate refuses to report when its input is **incomplete**, measured in
       **coverage artifacts** — the per-assembly Cobertura XML it actually consumes at
       `run-coverage.ps1:79-82` — not in TRX files, which Coverage never produces.
-- [ ] The threshold error stops claiming coverage when the cause is missing input. The
+- [x] The threshold error stops claiming coverage when the cause is missing input. The
       current text sends a reader hunting for a coverage regression that does not exist.
-- [ ] Timestamps are **not** used to prove freshness. With concurrent runs a file newer than
+- [x] Timestamps are **not** used to prove freshness. With concurrent runs a file newer than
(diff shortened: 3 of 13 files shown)


s205109:

s205109:
## Review findings addressed

Both findings from the PR Reviewer Guide were verified against the code and the running scripts. Both are incorrect, so no code changed.

| # | Finding | Verdict | Evidence |
|---|---|---|---|
| 1 | Broken `dotnet sln` call — passing a directory makes the command exit non-zero | **Pushback** | `dotnet sln <directory> list` is supported and exits 0 |
| 2 | False test failure — `$testExitCode` stays `$null` when every project passes | **Pushback** | `scripts/run-coverage.ps1:61` sets `$testExitCode = 0` before the loop |

### 1. Broken `dotnet sln` call — pushback

`dotnet sln` accepts a **directory** as well as a solution file. Given a directory it resolves the single solution file inside it. Run in this repository root:

```
$ dotnet sln <repo root> list
Project(s)
----------
AHKFlowApp.csproj
src\Backend\AHKFlowApp.API\AHKFlowApp.API.csproj
src\Backend\AHKFlowApp.Application\AHKFlowApp.Application.csproj
src\Backend\AHKFlowApp.Domain\AHKFlowApp.Domain.csproj
...
exit code 0
```

Two more independent proofs on this branch:

- `tests/CoverageInputCompleteness.Tests.ps1` case 4 calls `Get-AhkFlowCoverageProject -RepoRoot $repoRoot` and asserts the returned project list. It passes locally and in the `powershell-suites` CI job on this branch.
- `pwsh .\scripts\test-fast.ps1 -Mode Coverage` completed and printed line 94.6% / branch 82.6%. It could not have reached the report if this function threw.

The directory form is deliberate. It keeps the solution file name out of the helper, so renaming `AHKFlowApp.slnx` does not break coverage. It has one failure mode the file form does not — two solution files in the repository root — and there is exactly one (`AHKFlowApp.slnx`).

### 2. False test failure — pushback

The snippet in the finding starts below the initialization. `scripts/run-coverage.ps1:61` sets `$testExitCode = 0` before the `try` block that holds the loop:

```powershell
$previousSharedSqlConnectionString = $env:AHKFLOW_TEST_SQL_CONNECTION_STRING
$testExitCode = 0            # <- line 61, before the loop
try {
    ...
    foreach ($project in $coverageProject) {
        ...
        if ($LASTEXITCODE -ne 0) { $testExitCode = $LASTEXITCODE }
    }
}
```

That line is unchanged from before this PR, so there is no regression. When every project passes, `$testExitCode` is `0` and `if ($testExitCode -ne 0)` is `$false`. The full Coverage run on this branch confirms it: all projects passed, the script did not throw, and it printed the coverage numbers and the report paths.

The conditional assignment is intentional. It keeps the **first** non-zero exit code instead of letting a later passing project overwrite a failure with `0`, which is what an unconditional `$testExitCode = $LASTEXITCODE` inside a per-project loop would do.

---

**Gate:** not re-run — no code changed. The evidence above comes from the gate run on this exact commit: `powershell-suites` green in CI, and locally all 20 PowerShell suites, `dotnet build` Release with 0 warnings, `dotnet format --verify-no-changes`, and `-Mode Coverage` green.

**Deferred:** none.

These findings arrived as a single plain PR comment rather than review threads, so there was nothing to resolve.

PR #296, opened 2026-08-13
feat: guard reads link targets as write targets (backlog 083, redelivery)

s205109:
## What

Closes backlog 083. The agent worktree guard now reads the path a link **points to**, not
only the path it is created **at**.

Before this, a worktree session could create a link inside an allowed location aimed at a
file in the main checkout, then write through it. The write looked allowed, because the
guard only ever saw the allowed path.

## Why this PR exists

PR #291 merged early. It carried only the three `chore:` stage commits (pickup, 2-design,
3-plan) — the seven implementation commits never reached the remote, and the branch was
deleted after the merge. `origin/main` still has the item open at `backlog/083-*.md` with
no link handling in the guard. This PR delivers the work that #291 was supposed to.

## Spellings covered

| Spelling | Target read from |
|---|---|
| `New-Item -ItemType HardLink\|SymbolicLink\|Junction` | `-Target`, `-Value` |
| `ln`, `ln -s` | every operand |
| `mklink`, `/D`, `/H`, `/J` | link and target |
| `cp --link`, `cp --symbolic-link` | sources |
| `cmd /c ...`, `cmd //c "..."` | the inner command |

`New-Item` reads a target only behind an `ItemType` gate, so a plain file or directory
creation is unaffected. A link whose target is inside the session's own worktree stays
allowed.

## Merge note

`origin/main` moved while this sat undelivered (084 pipeline-bound sources, 085 item,
SSH.NET pin). Merged in and resolved two conflicts in
`scripts/agents/agent-worktree-guard.common.ps1`. Both were the same shape: 084 added a
pipeline-bound source check inline, this branch had extracted the interpreter scan into
`Get-AgentInterpreterInnerTarget`. Kept both — 084's check, then the shared helper, which
is a superset of the inline block it replaces (same depth cap and spec matching, plus the
unquoted-rest scan).

## Verification

Canonical five-step gate, all green:

- `dotnet format AHKFlowApp.slnx --verify-no-changes` — exit 0
- `pwsh .\scripts\test-fast.ps1 -Mode PowerShell` — 18/18 suites passed
- `pwsh .\scripts\test-fast.ps1 -Mode Coverage` — all thresholds met, line 94.6% / branch 82.6%
- `git diff --check origin/main...HEAD` — exit 0

`AgentWorktreeGuard.Tests.ps1`: 673 assertions, 0 failures. 084's `PipedFrom:` tests and
083's link tests both green in one tree.

### DIFF
--- a/backlog/086-guard-reads-the-link-kind-before-anchoring-a-relative-target.md
+++ b/backlog/086-guard-reads-the-link-kind-before-anchoring-a-relative-target.md
@@ -1,0 +1,79 @@
+# 086 - Guard reads the link kind before anchoring a relative target
+
+## Metadata
+
+- **Epic**: Agent guardrails
+- **Type**: Feature
+- **Interfaces**: none (agent git guard)
+- **Stage**: 0-intake
+
+## Summary
+
+The guard offers a relative link target three ways, because it does not know whether the command
+creates a symbolic link or a hard link. Reading the kind first would let each form keep only the
+anchor that matches how the operating system resolves it. That removes a refusal that stops
+ordinary work, without giving up any denial the guard makes today.
+
+## User story
+
+As an agent working in a worktree, I want a relative link that stays inside my own worktree to be
+allowed, so that I do not have to rewrite ordinary link commands to get past the guard.
+
+## Detail
+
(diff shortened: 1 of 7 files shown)


s205109:

PR #300, opened 2026-08-13
fix: hide watcher console window on worktree removal

s205109:
## What

The worktree removal watcher no longer flashes a console window.

`Win32_Process.Create` received no `ProcessStartupInformation`, so Windows created the
watcher with a visible console window. `-WindowStyle Hidden` is a PowerShell host argument,
so the host hid that window only after it started. The gap is what the user saw as a black
flash — measured at 135-369 ms. A full PowerShell suite run drives five removals, so it
flashed five times.

The spawn now passes a `Win32_ProcessStartup` instance with `ShowWindow = 0` (`SW_HIDE`),
which the system applies when it creates the window. The window is never shown.

## The backlog item's proposed fix does not work

Backlog 088 proposed `CreateFlags = 0x08000000` (`CREATE_NO_WINDOW`) and called
`ShowWindow = 0` "the weaker alternative". Measurement reversed that:

| Variant | Result |
|---|---|
| No startup information (old code) | Visible `PseudoConsoleWindow` 135-369 ms after the spawn |
| `ShowWindow = 0` | No visible window |
| `CreateFlags = 8` (`Detached_Process`) | `Create` returns 0, but the child never runs: a PowerShell host with no console does nothing |
| `CreateFlags = 0x08000000` (`CREATE_NO_WINDOW`) | `Create` returns **21**, invalid parameter, and creates no process |

`CREATE_NO_WINDOW` is not in the accepted `CreateFlags` values that the
[`Win32_ProcessStartup` documentation](https://learn.microsoft.com/windows/win32/cimwin32prov/win32-processstartup)
lists. `ShowWindow = 0` is the
[documented way to run a process hidden through WMI](https://learn.microsoft.com/windows/win32/wmisdk/wmi-tasks--processes).

Shipping `CREATE_NO_WINDOW` would have sent every removal down the `Start-Process` fallback,
which the session can kill. The reversal is recorded in the closed backlog item.

## Changes

- `scripts/worktree-powershell.common.ps1` — new `New-HiddenProcessStartup`. Returns `$null`
  when the CIM class is unavailable, so a removal that cannot hide its window still happens.
- `scripts/remove-worktree-local-dev.ps1` — passes `ProcessStartupInformation` when the
  helper returns an instance, and logs when it does not.
- `tests/WorktreeWatcherWindow.Tests.ps1` — new suite.
- `backlog/088-...md` — boxes ticked, moved to `backlog/done/`.

## Verification

The new suite spawns a probe through the same WMI call and polls `EnumWindows` for any
visible top-level window owned by it. The scan runs until the probe's host has started plus
500 ms, rather than for a fixed span, so a slow runner cannot produce a false pass. A probe
that never starts fails the suite.

Proven to bite: strip `ProcessStartupInformation` from the probe spawn and it fails with
`A visible window appeared for the spawned watcher after 172ms: PseudoConsoleWindow`.

Gate: build 17 projects 0 errors; `dotnet format --verify-no-changes` clean; 21/21 PowerShell
suites pass, `WorktreeRemoveHook` and `WorktreeMergedCleanup` included; coverage all
thresholds met, line 94.6%; `git diff --check main...HEAD` clean.

Closes backlog 088.

### DIFF
--- a/backlog/done/088-worktree-removal-watcher-flashes-a-console-window.md
+++ b/backlog/done/088-worktree-removal-watcher-flashes-a-console-window.md
@@ -5,8 +5,8 @@
 - **Epic**: Development process
 - **Type**: Tooling
 - **Interfaces**: none (scripts)
-- **Difficulty**: trivial
-- **Stage**: 0-intake
+- **Difficulty**: moderate
+- **Stage**: 9-ship
 
 ## Summary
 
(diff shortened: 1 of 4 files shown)


s205109:
Human visual check: two full `run-powershell-suites.ps1` runs, ten worktree removals in total, no window flash observed. That closes the acceptance criterion no test can prove.

PR #304, opened 2026-08-13
fix: guard reads the link kind before anchoring a relative target

s205109:
Backlog 086.

The agent worktree guard offers a relative link target three ways, because it does not know
whether the command creates a symbolic link or a hard link. Each anchor is correct for one kind
and wrong for the other, so keeping all three refuses ordinary work: `ln -s ../src/a.cs
deep/bait.md` stays inside the worktree, and the guard still denies it.

This branch reads the link kind first, then keeps only the anchors that match how the operating
system resolves that kind.

- A symbolic link keeps the two link-relative anchors.
- A hard link keeps the working-directory anchor.
- A junction, or a kind the guard cannot read, keeps all three and stays fail-closed.

Stage: 1-pickup

### DIFF
--- a/backlog/086-guard-reads-the-link-kind-before-anchoring-a-relative-target.md
+++ b/backlog/086-guard-reads-the-link-kind-before-anchoring-a-relative-target.md
@@ -5,7 +5,8 @@
 - **Epic**: Agent guardrails
 - **Type**: Feature
 - **Interfaces**: none (agent git guard)
-- **Stage**: 0-intake
+- **Stage**: 1-pickup
+- **Difficulty**: moderate
 
 ## Summary
 
--- a/backlog/091-guard-option-gates-accept-the-abbreviations-the-real-tools-accept.md
+++ b/backlog/091-guard-option-gates-accept-the-abbreviations-the-real-tools-accept.md
@@ -1,0 +1,71 @@
+# 091 - Guard option gates accept the abbreviations the real tools accept
+
+## Metadata
+
+- **Epic**: Agent guardrails
+- **Type**: Bug
+- **Interfaces**: none (agent git guard)
+- **Stage**: 0-intake
+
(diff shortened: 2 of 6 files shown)


s205109:
Review round 2 — the `-S` hole is fixed in `69c5f98`.

`-S` is `--suffix` and takes a value, so `ln -S -s ../README.md deep/bait.md` sets the suffix to
`-s` and creates a HARD link. The kind reader read that `-s` as the symbolic flag, dropped the
as-written anchor, and allowed a hard link into the main checkout.

The fix generalises what the `-t` fix did for one option. The reader now walks a short cluster
letter by letter and stops at the first option that needs a value; the rest of the token, or the
whole next token, is that value and is no longer read as an option. The long spellings
`--suffix X` and `--target-directory X` consume their value the same way, while `--suffix=X`
does not.

Pinned at `tests/AgentWorktreeGuard.Tests.ps1`: `ln -S -s ../README.md deep/bait.md` denies,
`ln -S -s src/a.cs deep/bait.md` allows, plus nine kind-reader cases covering `-Ss`, `-St out`,
`-S .bak -s`, `--suffix -s`, and `--suffix=.bak -s`.

**Gate note.** A reviewer's run reported 21 passed and 1 failed, with
`WorktreePlansSymlink.Tests.ps1` failing. On this machine the same commit reports 22 of 22
passed. The cause is environmental: that test looks up `bash` and `rg` through PowerShell, then
runs the script under `bash`. Where `bash` resolves to the WSL launcher, the script runs with the
Linux PATH and cannot see the Windows `rg`. This branch touched neither that test nor the script
it exercises. Filed as backlog 092.

Backlog 091 holds the two deferred findings from round 1: `cp --sy` and `New-Item -ItemType Sym`
both walk past exact-match gates that predate this branch.

PR #308, opened 2026-08-14
fix: guard option gates accept abbreviated spellings

s205109:
## Problem

Two gates in the agent worktree guard compared an option or a type name for exact
equality. The real tools accept shorter spellings, so a link command walked past its
gate and the guard never scanned the path the link pointed at. Both holes let a link
reach into the main checkout.

Found in review of #304 (backlog 086). Both predate it.

**Gate 1 — the `cp` link flag.** `Test-AgentGuardHasLinkFlag` matched `--link` and
`--symbolic-link` with `-ceq`, plus a short cluster with `-cmatch '^-[a-zA-Z]*[ls]'`.
The cluster pattern cannot match a long option: after the first `-`, `[a-zA-Z]*`
cannot cross the second `-`, so `[ls]` is tested against `-` and fails. `cp --sy a b`
therefore matched nothing, the link branch never ran, and only the destination was
reported.

**Gate 2 — the `New-Item` item type.** `Get-AgentNewItemLinkTarget` tested the item
type against a literal list with `-notcontains`. The FileSystem provider does not
match exactly, so `New-Item -ItemType Sym` creates a real symbolic link and the guard
read no target for it. The kind switch made the same assumption, so an abbreviated
item type would also have read as an unknown kind.

## What the provider really does

The backlog item asked for this to be proved against the running provider rather than
copied from memory. The match is **not** a prefix match. It is a wildcard match of
`<token>*`, without case, against a fixed list tried in a fixed order.

Proved by running `New-Item` under `pwsh 7.6.4` and reading back `LinkType`:

| Item type written | Result |
|---|---|
| `Sym`, `S` | SymbolicLink |
| `hardl`, `H` | HardLink |
| `Junc`, `j` | Junction |
| `Fi` | a plain file |
| `D`, `container`, `cont` | a directory |
| `*link` | SymbolicLink |
| `?ardlink` | HardLink |
| `[sh]ymboliclink` | SymbolicLink |
| `bogus`, `directoryx` | an error, nothing created |

The order was settled by patterns that match two names at once: `[dh]*` makes a
directory, `[fs]*` and `*i*l` make a file, `[js]*` makes a symbolic link, and `[jh]*n`
takes the junction path. So the order is `directory`, `container`, `file`,
`symboliclink`, `junction`, `hardlink`.

One more fact the fix needs: a malformed pattern throws.
`'symboliclink' -like '[sh*'` raises `WildcardPatternException`. An uncaught throw
escapes into the entrypoint's catch, and that catch **allows** the write.

## Change

- `Test-AgentGuardHasLinkFlag` keeps its cluster test and reads long options through
  `Test-AgentLongOptionPrefix`, the helper `Get-AgentLinkKind` already uses.
- New `Get-AgentNewItemType` resolves an item-type token the way the provider does,
  and returns `$null` when it resolves to none. `$null` is the fail-closed answer.
- `Get-AgentNewItemLinkTarget` reads the link value for every result except
  `Directory` and `File`, and maps `SymbolicLink` to `Symbolic` and `HardLink` to
  `Hard` when it chooses the anchor.
- `$script:AgentGuardLinkItemTypes` had no other caller and is gone.

Both matches over-report rather than under-report. An ambiguous abbreviation such as
`--s`, which real `cp` refuses to run, is accepted, because every `cp` operand is a
path anyway and under-reporting is the failure that matters.

## Verification

`pwsh ./scripts/test-fast.ps1 -Mode PowerShell` — 22 of 22 suites pass.

New coverage in `tests/AgentWorktreeGuard.Tests.ps1`: 2 kind cases, 12 write-target
cases, 8 decision cases. Three decision rows split on the kind, so an abbreviation
that read as `Unknown` would fail them:

| Command | Expected |
|---|---|
| `New-Item -ItemType Sym -Path deep\bait.md -Target ..\src\a.cs` | Allow |
| `New-Item -ItemType hardl -Path deep\bait.md -Target ..\README.md` | Deny |
| `New-Item -ItemType bogus -Path deep\bait.md -Target ..\README.md` | Deny |

## Out of scope

- The anchoring rules for a relative target, which backlog 086 settled.
- The junction anchor, which is still unproved and still fails closed.

Closes backlog 091.

### DIFF
--- a/backlog/093-guard-tokenizer-splits-powershell-backtick-and-parenthesis-dropping-the-command-leaf.md
+++ b/backlog/093-guard-tokenizer-splits-powershell-backtick-and-parenthesis-dropping-the-command-leaf.md
@@ -1,0 +1,80 @@
+# 093 - Guard tokenizer splits PowerShell backtick and parenthesis, dropping the command leaf
+
+## Metadata
+
+- **Epic**: Agent guardrails
+- **Type**: Bug
+- **Interfaces**: none (agent git guard)
+- **Stage**: 0-intake
+
+## Summary
+
+The guard tokenizer treats a backtick and a parenthesis as command separators, because they are
+separators in bash. They are not separators in PowerShell: a backtick escapes the character behind
+it, and parentheses group an expression. A PowerShell command that holds either one is cut into
+segments whose leading word is no longer the command, so the guard reads no write target for it
+and allows the write.
+
+## User story
+
+As a person who owns the main checkout, I want the guard to read a PowerShell command as
+PowerShell, so that an agent cannot reach a protected file by putting a backtick in front of it.
+
+## Detail
+
+The separator list is at `scripts/agents/agent-worktree-guard.common.ps1:717-718`. It puts
+`` ` ``, `(`, and `)` beside `;`, `&`, and `|`.
+
+Every command below was run through `Invoke-AgentGuardPolicy` with the working directory set to a
+managed worktree and the protected root set to the main checkout. Each one returned **Allow**, with
+no write targets and no `Unresolved` flag. The same command without the backtick returns Deny:
+
+```powershell
+# Deny, as it should be.
+Set-Content -Path <main>\README.md -Value x
+
+# Allow. Segment 1 is 'Set-Content -Path'; segment 2 leads with the path, so its leading word is
+# not a command the guard knows, and no target is read.
+Set-Content -Path `<main>\README.md -Value x
+Set-Content -Path ('<main>' + '\README.md') -Value x
+Remove-Item -LiteralPath `<main>\README.md
+rm `<main>/README.md
+```
+
+The same shape defeats the link rules, which is where this was found:
+
+```powershell
+# Both create a real symbolic link into the main checkout. Both return Allow.
+New-Item -ItemType `Sym -Path <managed>\bait.md -Target <main>\README.md
+New-Item -ItemType ('Sym'+'bolicLink') -Path <managed>\bait.md -Target <main>\README.md
+```
+
+The link examples are not a link-rule bug. `Get-AgentNewItemLinkTarget` never receives the item
+type, because the tokenizer removed it from the segment before the reader ran. The `Set-Content`
+and `rm` examples above carry no item type and no link, and they fail the same way.
+
+## Acceptance criteria
+
+- [ ] A PowerShell command holding a backtick is read as one command, or refused
+- [ ] A PowerShell command holding a parenthesised expression is read as one command, or refused
+- [ ] `rm` and `Set-Content` with a backtick before a main-checkout path report Deny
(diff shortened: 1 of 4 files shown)


s205109:
## Review findings addressed

| # | Finding | Verdict | Fix |
|---|---|---|---|
| 1 | Backtick/paren item-type bypass | DEFER | Filed as backlog 093 (rides along in this PR) |
| 2 | Shell-expanded `cp` options bypass the gate | CONFIRMED | `Get-AgentShellLiteralHead`, `common.ps1:1200` |
| 3 | No direct tests for `Get-AgentNewItemType` | CONFIRMED | New 23-row block, `Tests.ps1:2083` |
| 4 | `D`/`cont` rows do not distinguish `$null` | CONFIRMED | Both carry `-Value`, `Tests.ps1:2335` |
| 5 | Provider transcript too long in code | CONFIRMED | Cut to rule, precedence, and plan pointer, `common.ps1:2311` |
| 6 | `''` resolves to Directory, provider makes a file | CONFIRMED | Resolves to `File`, `common.ps1:2327` |

All six were reproduced against the running guard before triage.

### Finding 2

The guard reads a command as typed, not as the shell expanded it. `cp --s\y` reaches
cp as `--sy`, and `cp --l${x}in` reaches it as `--lin`. Each token is now cut at the
first character a shell acts on, and the head is what gets matched.

The two option shapes then part company. A long option is decided by its head, so
`--sparse=$WHEN` stays a plain copy and `--target-directory=out\sub` stays a
destination. A short cluster carries its letters one after another, so `-a\l` hides
the very letter that names the link and the head `-a` proves nothing. Any cut in a
short cluster is treated as a link flag instead.

One step past the finding: `cp -a\l` defeats the same gate the same way, and fixing
only the long-option half would have been knowingly incomplete. It failed red with
the others and now passes.

### Finding 1 - why it is deferred, not fixed

Real, and reproduced. It is not a link-rule defect. Every command below also returns
**Allow**, with no item type and no link anywhere:

```
rm `<main>/README.md
Set-Content -Path `<main>\README.md -Value x
Set-Content -Path ('<main>' + '\README.md') -Value x
Remove-Item -LiteralPath `<main>\README.md
```

The cause is the tokenizer at `agent-worktree-guard.common.ps1:717-718`, which treats
a backtick and parentheses as command separators. That is right for bash and wrong for
PowerShell, where a backtick escapes the character behind it and parentheses group an
expression. `Get-AgentNewItemLinkTarget` never receives the item type, because the
tokenizer removed it from the segment before the reader ran.

It predates this branch. `git diff main...HEAD` puts the hunks at `1186-1209`, `1610`,
and `2243-2350`, and the tokenizer is untouched. The fix has to know which shell a
command is bound for, or refuse the ambiguity, and that is its own change rather than a
rider on this one. Backlog 093 carries the reproduction and the acceptance criteria.

### Gate

`pwsh ./scripts/test-fast.ps1 -Mode PowerShell` - 22 of 22 suites pass.

